PR #14, opened 2023-03-17
a possible extension of RDF's Simple semantics supporting quoted triples

pchampin:
mostly for discussion at this point

(I already discussed it briefly with @doerthe and @Antoine-Zimmermann )

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/14.html" title="Last updated on Mar 24, 2023, 8:14 AM UTC (26b68f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/14/a46453d...26b68f7.html" title="Last updated on Mar 24, 2023, 8:14 AM UTC (26b68f7)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -96,6 +96,11 @@
     table, td, th { border:1px solid black; }
     caption { font-weight: bold; text-align: left ; }
     code {color: #ff4500;}  /* Old W3C Style */
+
+    .new {
+      background-color: rgba(255, 128, 0, 0.2);
+      color: darkred;
+    }
   </style>
 </head>
 <body>
@@ -252,7 +257,7 @@
       so that each triple in a graph is considered to be a subgraph.
       A <dfn>proper subgraph</dfn> is a proper subset of the triples in the graph.</p>
 
-    <p>A <dfn>ground</dfn> RDF graph is one that contains no blank nodes.</p>
+    <p><span class="new">A <dfn>ground</dfn> term is either a <a>name</a> or a quoted triple containing only ground terms.</span> A ground RDF graph is one that contains no blank nodes (only ground terms).</p>
 
     <p>Suppose that M is a functional mapping from a set of blank
       nodes to some set of literals, blank nodes and IRIs. Any graph obtained
@@ -400,6 +405,7 @@
               set of sets of pairs &lt; x, y &gt; with x and y in IR .</p>
         <p>4. A mapping IS from IRIs into (IR union IP)</p>
         <p>5. A partial mapping IL from literals into IR </p>
+        <p class="new">6. A mapping IT from <a>ground</a> quoted triples into IR</p>
        </td>
     </tr>
   </table>
@@ -457,10 +463,14 @@
       </tr>
 
       <tr>
+        <td class="new semantictable">if E is a <a>ground</a> quoted triple then I(E) = IT(E)</td>
+      </tr>
+
+      <tr>
         <td class="semantictable">if E is a ground triple `s p o.`
           then I(E) = true if<br/>
-          I(p) is in IP and the pair &lt;I(s),I(o)&gt;
-          is in IEXT(I(p))<br/>
+          I'(p) is in IP and the pair &lt;I'(s),I'(o)&gt;
+          is in IEXT(I'(p))<br/>
           otherwise I(E) = false.</td>
       </tr>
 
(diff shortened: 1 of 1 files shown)


pchampin:
the only changes for the moment are in the 'Simple interpretation' section, and are highlighted to be easy to spot.

pfps:
I don't think it is a good idea to have suggestions on technical changes be made in pull requests.  My view is that technical changes should be suggested instead in issues or on the working group email (I prefer email) and discussion should happen there, not in a pull request.  

If someone wants to create a version of a document to support their suggestion then this can be done without creating a pull request.

TallTed:
> I don't think it is a good idea to have suggestions on technical changes be made in pull requests.

PRs are a good way to collaborate on textual edits, technical or otherwise.

Such PRs should *generally* come after some discussion of the proposed change in an issue, but this need not delve deeply into the technicalities nor exact text changes.

It makes sense to me to use PRs to propose document revisions, in large part because the document markup is often important to full comprehension of those proposals, and the PR Preview is only available via PR -- not via email, issue, or other discussion medium.

spec/index.html
```diff
@@ -495,25 +505,42 @@ <h3>Blank nodes</h3>
...
-      Define the mapping [I+A] to be I on <a>names</a>,
-      and A on blank nodes on the set: [I+A](x)=I(x) when x is a <a>name</a>
-      and [I+A](x)=A(x) when x is a blank node;
-      and extend this mapping to triples and RDF graphs using the rules given above for ground graphs.
-      Then the semantic conditions for an RDF graph are:</p>
+    <p>Suppose that</p>
+    <ul>
+      <li>I is a simple interpretation,</li>
+      <li>A is a mapping from a set N of blank nodes to the universe IR of I, and</li>
+      <li class="new">B is a mapping from a subset of N to the set of <a>ground</a> terms, such that for every x in that subset, I(B(x)) = A(x).</li>
```
TallTed, line 512:
```suggestion
    <p>Suppose the following:</p>
    <ul>
      <li>I is a simple interpretation</li>
      <li>A is a mapping from a set N of blank nodes to the universe IR of I</li>
      <li class="new">B is a mapping from a subset of N to the set of <a>ground</a> terms, such that for every x in that subset, I(B(x)) = A(x)</li>
```

spec/index.html
```diff
@@ -495,25 +505,42 @@ <h3>Blank nodes</h3>
...
+      <li>B is extended to names and quoted triples:
+        <ul>
+          <li>if x is a <a>name</a>, B(x) = x</li>
+          <li>if x is a quoted triple (s, p, o), B(x) = (B(s), B(p), B(o))</li>
```
TallTed, line 516:
```suggestion
          <li>If x is a <a>name</a>, B(x) = x</li>
          <li>If x is a quoted triple (s, p, o), B(x) = (B(s), B(p), B(o))</li>
```

spec/index.html
```diff
@@ -495,25 +505,42 @@ <h3>Blank nodes</h3>
...
+      </li>
+    </ul>
+
+    <p>Define the mapping [I+A+B] as follows:</p>
+    <ul>
+      <li>if x is a <a>ground</a> term, [I+A+B](x)=I(x)</li>
+      <li>if x is a blank node, [I+A+B](x)=A(x)</li>
+      <li class="new">if x is a non-<a>ground</a> quoted triple, then [I+A+B](x)=I(B(x)) </li>
+      <li>and extend this mapping to triples and RDF graphs using the rules given above for ground graphs.</li>
+    </ul>
```
TallTed, line 527:
```suggestion
    <ol>
      <li>If x is a <a>ground</a> term, [I+A+B](x)=I(x)</li>
      <li>If x is a blank node, [I+A+B](x)=A(x)</li>
      <li class="new">If x is a non-<a>ground</a> quoted triple, then [I+A+B](x)=I(B(x)) </li>
      <li>Extend this mapping to triples and RDF graphs using the rules given above for ground graphs.</li>
    </ol>
```

spec/index.html
```diff
@@ -456,11 +462,15 @@ <h2>Simple Interpretations</h2>
         <td class="semantictable">if E is an IRI then I(E) = IS(E)</td>
       </tr>
 
+      <tr>
+        <td class="new semantictable">if E is a <a>ground</a> quoted triple then I(E) = IT(E)</td>
```
TallTed, line 466:
```suggestion
        <td class="semantictable">if E is an IRI, then I(E) = IS(E)</td>
      </tr>

      <tr>
        <td class="new semantictable">if E is a <a>ground</a> quoted triple, then I(E) = IT(E)</td>
```

spec/index.html
```diff
@@ -456,11 +462,15 @@ <h2>Simple Interpretations</h2>
...
+      </tr>
+
       <tr>
         <td class="semantictable">if E is a ground triple `s p o.`
           then I(E) = true if<br/>
```
TallTed, line 471:
```suggestion
        <td class="semantictable">if E is a ground triple `s p o.`,
          then I(E) = true if<br/>
```

spec/index.html
```diff
@@ -456,11 +462,15 @@ <h2>Simple Interpretations</h2>
...
-          I(p) is in IP and the pair &lt;I(s),I(o)&gt;
-          is in IEXT(I(p))<br/>
+          I'(p) is in IP and the pair &lt;I'(s),I'(o)&gt;
+          is in IEXT(I'(p))<br/>
```
TallTed, line 473:
```suggestion
          is in IEXT(I'(p));<br/>
```

spec/index.html
```diff
@@ -456,11 +462,15 @@ <h2>Simple Interpretations</h2>
...
           otherwise I(E) = false.</td>
```
TallTed, line 474:
```suggestion
          otherwise, I(E) = false.</td>
```

spec/index.html
```diff
@@ -495,25 +505,42 @@ <h3>Blank nodes</h3>
...
+    <p>Then the semantic conditions for an RDF graph are:</p>
 
     <table>
       <caption>Semantic condition for blank nodes.</caption>
       <tbody>
         <tr>
-          <td class="semantictable">If E is an RDF graph then I(E) = true if [I+A](E) =
-          true for some mapping A from the set of blank nodes in E to IR, otherwise
+          <td class="semantictable">If E is an RDF graph then I(E) = true if [I+A+B](E) =
+          true for some mappings A <span class="new">and B as defined above</span>, otherwise
```
TallTed, line 536:
```suggestion
          <td class="semantictable">If E is an RDF graph, then I(E) = true if [I+A+B](E) =
          true for some mappings A <span class="new">and B as defined above</span>; otherwise,
```

pfps:
I'm not sure in what respect this proposal is minimal.  There is a considerable amount of machinery here.  A truly minimal proposal might treat quoted triples just like a new datatype with a lexical to value mapping that maps into a triple of strings.   A different minimal proposal might treat quoted triples as events.  To determine what a minimal proposal should look like requires determining what quoted triples are supposed to be used for.

pchampin:
> I'm not sure in what respect this proposal is minimal. 

Granted. I updated the title of the PR accordingly. The goal of this proposal is to be essentially equivalent to the semantics specified in the CG report through the unstar mapping (although in some corner cases, it might differ).

> Discussing quoted triples in semantics really depends on the core concepts being defined.

Agreed. The underlying assumption here is that the WG would adopt the changes made by the CG to the abstract syntax.

spec/index.html
pchampin: I would rather keep it that way for the moment, as this is the same style as other similar definitions inherited from the RDF 1.1 spec. I generally agree that it improves readability, but for the sake of consistency, we should apply this change globally.

pchampin: I see you actually updated them all. Fine by me then.

pfps:
An extract of PR showing only the parts relevant for the semantics would make it much easier to see what is going on and make it easier to compare with other proposals.

pfps:
I'm taking the "needs discussion" off this PR so that it won't show up in the list of issues imminently needing discussion.    The label should be added if this PR ever gets out of a draft state.

pfps:
As far as I can tell, this semantics has the following characteristic.

Suppose we have a class for triples.  Then

:john :said _:s .
_:s rdf:type rdf:Triple .

does not entail

:john :said << _:a _:b _:c >> .

pfps:
Suppose we have a class for triples.  Then

:john :said _:s .
_:s rdf:type rdf:Triple .

does not entail

:john :said << _:a _:b _:c >> .

because there might not be any ground triple that is mapped by IT to the domain element that John said.

afs:
Is this PR superseded?

pchampin:
> Is this PR superseded?

Very much so! Closing it to avoid any confusion.

PR #15, opened 2023-03-17
remove 'typed' and 'datatyped' as adjectives on literal

pfps:
RDF 1.0 had untyped and typed literals.   RDF 1.1 has only typed literals but some occurrences of ''typed literal" and "datatyped literal" remained in RDF semantics.  They are mostly harmless so it should be considered an editorial change to remove the unnecessary adjectives.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/15.html" title="Last updated on Apr 1, 2023, 5:04 PM UTC (5bb5e02)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/15/9f41603...5bb5e02.html" title="Last updated on Apr 1, 2023, 5:04 PM UTC (5bb5e02)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -242,7 +242,7 @@
       or other components of RDF syntax.
       Some cases are illustrated by node-arc diagrams showing the graph structure directly.</p>
 
-    <p>A <dfn>name</dfn> is any IRI or literal. A typed literal contains
+    <p>A <dfn>name</dfn> is any IRI or literal. A literal contains
       two <a>names</a>: itself and its internal type
       IRI. A <dfn>vocabulary</dfn> is a set of <a>names</a>.</p>
 
@@ -779,7 +779,7 @@
     The <a>value space</a> of <code>rdf:langString</code> is the set of all pairs of a string with a language tag.
     The semantics of literals with this as their type are given below.</p>
 
-  <p>RDF literal syntax allows any IRI to be used in a typed literal,
+  <p>RDF allows any IRI to be used in a literal,
     even when it is not <a>recognized</a> as referring to a datatype.
     Literals with such an "unknown" datatype IRI,
     which is not in the set of <a>recognized</a> datatypes,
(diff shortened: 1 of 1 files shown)


gkellogg:
It might be good to have a way term to distinguish language-tagged strings from other literals. I've seen the term "datatyped literal" used to distinguish between the two. Previously, there would have been three types of literals: "plain", "datatyped", and "language tagged".

If a term is removed, we should probably try to keep some anchor (e.g., `<span id="dfn-..."></span>`) so that references to the old term continue to resolve.

TallTed:
> It might be good to have a way term to distinguish language-tagged strings from other literals.

We have a way: [langstring](http://www.w3.org/1999/02/22-rdf-syntax-ns#langString) vs [string](http://www.w3.org/2001/XMLSchema#string).

gkellogg:
`language-tagged string` describes a literal with a `language-tag` and datatype `rdf:langString`. Other literals have a datatype, but no language-tag, and the datatype must not be `rdf:langString`. There is a component which is a string, but I would not say that `string` describes literals that are not language-tagged.

pfps:
@gkellogg Is your request to have a term that describes all literals except language-tagged strings?  The place to introduce that is almost certainly in RDF Concepts, where the other syntactic categories are defined, not in RDF Semantics.

spec/index.html
```diff
@@ -778,7 +778,7 @@ <h2>Literals and datatypes</h2>
     The <a>value space</a> of <code>rdf:langString</code> is the set of all pairs of a string with a language tag.
     The semantics of literals with this as their type are given below.</p>
 
-  <p>RDF literal syntax allows any IRI to be used in a typed literal,
+  <p>RDF literal syntax allows any IRI to be used in a literal,
```
gkellogg, line 781: Other than for the language-tagged string syntax, which is a literal that does not allow an IRI to be specified. Perhaps just add a parenthetical.

gkellogg (approved):
Other than the slight inaccuracy I noted, this looks good to me.

pfps:
The RDF Semantics document uses syntax to refer to the abstract syntax of RDF, where every literal does indeed have a datatype IRI.  But it would be better to just say "RDF" instead of "RDF literal syntax".

PR #27, opened 2023-04-25
fix up definition anchors for rules and links to RDF Concepts

pfps:
needed for citations from SPARQL Entailment Regimes
also fix links to RDF Concepts

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/27.html" title="Last updated on May 4, 2023, 4:57 PM UTC (b780a1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/27/a45c9a0...b780a1f.html" title="Last updated on May 4, 2023, 4:57 PM UTC (b780a1f)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -997,7 +997,7 @@
       For example, all the RDF axioms are true in every <a>RDF interpretation</a>,
       and so are <a>RDF entail</a>ed by the empty graph, contradicting <a>interpolation</a> for RDF entailment. </p>
 
-    <section id="rdf_entailment_patterns" class="informative" data-dfn-for="RDF entailment patterns">
+    <section id="rdf_entailment_patterns" class="informative">
       <h4>Patterns of RDF entailment (Informative)</h4>
 
       <p>The last semantic condition in the above table gives the following entailment pattern for <a>recognized</a> datatype IRIs:</p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -995,7 +995,7 @@
       For example, all the RDF axioms are true in every <a>RDF interpretation</a>,
       and so are <a>RDF entail</a>ed by the empty graph, contradicting <a>interpolation</a> for RDF entailment. </p>
 
-    <section id="rdf_entailment_patterns" class="informative">
+    <section id="rdf_entailment_patterns" class="informative" data-dfn-for="RDF entailment patterns">
       <h4>Patterns of RDF entailment (Informative)</h4>
 
       <p>The last semantic condition in the above table gives the following entailment pattern for <a>recognized</a> datatype IRIs:</p>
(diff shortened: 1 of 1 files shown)


timothee-haudebourg (approved):

spec/index.html
```diff
@@ -1331,7 +1331,7 @@ <h3>RDFS entailment</h3>
 
     <p>where aaa is an IRI, are true in all RDFS interpretations.</p>
 
-    <section id="rdfs_patterns" class="informative">
+    <section id="rdfs_patterns" class="informative" data-dfn-for="RDFS entailment patterns">
```
gkellogg, line 1334: The data-dfn-for would seem to be intended to add the value as a prefix to the rule definition anchors, but it doesn’t seem to actually changed anything.

gkellogg: I can't really investigate right now, but a question on spec-prod@w3.org describing what you're trying to do might get some helpful feedback. I've always used data-dfn-for in conjunction with WebIDL definitions, for instance in JSON-LD API.

pfps: It is supposed to work, according to Tab Atkins.

gkellogg:
IIRC, in ReSpec, `data-dfn-for` really needs the "for" bit to be a WebIDL interface, and as there is none, it may not be doing what you expect. This is different in Bikeshed, where that isn't necessary.

You could try raising an issue on GitHub.com/w3c/respec and see how they resolve it.

Also, I think you can also use the `id` attribute on the `dfn` to give it a specific fragment identifier, for example `<dfn id="dfn-rdfs-entailment-rdfs1">...</dfn>` to override the default identifier issued; I think @afs had an issue like this in SPARQL Query.

pfps:
@gkellogg @TallTed I was hoping to have a common data-dfn-for.   The replies by Tab Atkins Jr. and Marcos Caceres (see https://lists.w3.org/Archives/Public/spec-prod/2023AprJun/0006.html) seem to indicate that this is possible.   But then there is https://github.com/w3c/respec/issues/4415 seems to indicate otherwise.    I'm thus unclear as to what question I should be asking.

gkellogg:
Not sure if Marcos considers that a bug or not, but as there are a relatively small number of definitions, putting `data-dfn-for` on the `dfn` defining the term seems like a place to start. You can look at the resulting anchor to see if implements what you expect, and if your references work as expected. Note that there is also a `data-ink-for` to use on anchors, but that might not be necessary using the native syntax.

TallTed:
s/[`data-ink-for`](https://github.com/w3c/rdf-semantics/pull/27#issuecomment-1526385781)/`data-link-for`/, I think

pfps:
@gkellogg @TallTed One problem is that the common directory isn't being found by Firefox on my local machine so I don't know whether the problems I am seeing locally are caused by this or not.   Is there a way to have Firefox access this file?

gkellogg:
I run `python -m http.server` from a parent directory and open via localhost, which should see the common files and images. For example `http://localhost:8000/rdf-semantics/spec/index.html`.

pfps:
Thanks.

PR #30, opened 2023-05-07
improve display on mobile phones

domel:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/30.html" title="Last updated on May 18, 2023, 10:09 PM UTC (e0c1a8a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/30/13277a8...e0c1a8a.html" title="Last updated on May 18, 2023, 10:09 PM UTC (e0c1a8a)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -97,8 +97,10 @@
     caption { font-weight: bold; text-align: left ; }
     code { color: #ff4500; }  /* Old W3C Style */
     table { word-break: normal; overflow-wrap: anywhere !important; }
+    table tr td:first-child { overflow-wrap: break-word; }
     @media (max-width: 550px) {
-        table th, table td { font-size: 12px; padding: 3px 4px;}
+        table th, table td { font-size: 14px; padding: 3px 4px;}
+        table td.semantictable, table td.ruletable { font-size: 14px;}
         .othertable {padding: 4px;}
         img {max-width: 85vw;}
     }  
--- a/spec/index.html
+++ b/spec/index.html
@@ -52,11 +52,11 @@
   }
 
   .technote::before {
-      content:    "Technical Note";
-      display:    block;
+      content: "Technical Note";
+      display: block;
       width:  150px;
       margin: -1.5em 0 0.5em 0;
-      font-weight:    bold;
+      font-weight: bold;
       border: 1px solid #cff6d9;
       background: #eff;
       padding:    3px 1em;
@@ -64,22 +64,22 @@
 
 
   .changenote {
-      font-size:small;
+      font-size: small;
       margin: 1em 0em 0em;
-      padding:    1em;
+      padding: 1em;
       border: 2px solid #cff6d9;
       background: #ffddfe;
   }
 
   .changenote::before {
-      content:    "Change Note";
-      display:    block;
+      content: "Change Note";
+      display: block;
       width:  150px;
       margin: -1.5em 0 0.5em 0;
-      font-weight:    bold;
+      font-weight: bold;
       border: 1px solid #cff6d9;
       background: #ffddef;
-      padding:    3px 1em;
+      padding: 3px 1em;
   }
 
 
@@ -92,10 +92,11 @@
       border: 1px solid black
   }
 
-    table { border-collapse:collapse; }
-    table, td, th { border:1px solid black; }
+    table { border-collapse: collapse; }
+    table, td, th { border: 1px solid black; }
     caption { font-weight: bold; text-align: left ; }
-    code {color: #ff4500;}  /* Old W3C Style */
+    code { color: #ff4500; }  /* Old W3C Style */
+    table { word-break: break-all; max-width: 80vw; }
   </style>
 </head>
 <body>
--- a/spec/index.html
+++ b/spec/index.html
@@ -40,7 +40,7 @@
   <style>
   .semantictable {background-color: #FFFFAA; padding:0.5em;}
   .ruletable {background-color: #DDDDFF; padding:0.5em;}
-  .othertable {background-color: #FDFDFD; padding:0.5em;}
+  .othertable {background-color: #FDFDFD; padding:0.2em;}
   .tabletitle {font-size: small; font-weight: bolder;}
 
   .technote {
--- a/spec/index.html
+++ b/spec/index.html
@@ -96,7 +96,12 @@
     table, td, th { border: 1px solid black; }
     caption { font-weight: bold; text-align: left ; }
     code { color: #ff4500; }  /* Old W3C Style */
-    table { word-break: normal; overflow-wrap: anywhere; max-width: 80vw; }
+    table { word-break: normal; overflow-wrap: anywhere !important; max-width: 80vw; }
+    @media (max-width: 850px) {
+        table th, table td { font-size: 12px; padding: 3px 4px;}
+        .othertable {padding: 4px;}
+        img {max-width: 80vw;}
+    }  
   </style>
 </head>
 <body>
--- a/spec/index.html
+++ b/spec/index.html
@@ -96,11 +96,11 @@
     table, td, th { border: 1px solid black; }
     caption { font-weight: bold; text-align: left ; }
     code { color: #ff4500; }  /* Old W3C Style */
-    table { word-break: normal; overflow-wrap: anywhere !important; max-width: 80vw; }
+    table { word-break: normal; overflow-wrap: anywhere !important; }
     @media (max-width: 850px) {
         table th, table td { font-size: 12px; padding: 3px 4px;}
         .othertable {padding: 4px;}
-        img {max-width: 80vw;}
+        img {max-width: 85vw;}
     }  
   </style>
 </head>
--- a/spec/index.html
+++ b/spec/index.html
@@ -96,7 +96,7 @@
     table, td, th { border: 1px solid black; }
     caption { font-weight: bold; text-align: left ; }
     code { color: #ff4500; }  /* Old W3C Style */
-    table { word-break: break-all; max-width: 80vw; }
+    table { word-break: normal; overflow-wrap: anywhere; max-width: 80vw; }
   </style>
 </head>
 <body>
--- a/spec/index.html
+++ b/spec/index.html
@@ -97,7 +97,7 @@
     caption { font-weight: bold; text-align: left ; }
     code { color: #ff4500; }  /* Old W3C Style */
     table { word-break: normal; overflow-wrap: anywhere !important; }
-    @media (max-width: 850px) {
+    @media (max-width: 550px) {
         table th, table td { font-size: 12px; padding: 3px 4px;}
         .othertable {padding: 4px;}
         img {max-width: 85vw;}


gkellogg (changes requested):
Looks like some markup problems; `dn` inside `dfn`.

pfps:
There are lots of changes that appear to be formatting.   They should be separated out into a separate PR.

pfps:
What does this actually do?

spec/index.html
```diff
@@ -92,10 +92,11 @@
...
   }
 
-    table { border-collapse:collapse; }
-    table, td, th { border:1px solid black; }
+    table { border-collapse: collapse; }
+    table, td, th { border: 1px solid black; }
     caption { font-weight: bold; text-align: left ; }
-    code {color: #ff4500;}  /* Old W3C Style */
+    code { color: #ff4500; }  /* Old W3C Style */
+    table { word-break: break-all; max-width: 80vw; }
```
TallTed, line 99:
```suggestion
    table { word-break: break-word; max-width: 80vw; }
```

TallTed:
I see only CSS changes — whitespace on several lines (looks like no effect on the stylesheet), and addition of one line (99):
```css
table { word-break: break-all; max-width: 80vw; }
```
I think that should be `break-word` instead of `break-all`, as discussed in another PR, and have no problem with the whitespace cleanup.

TallTed:
@gkellogg's concerns came from the Echidna Auto-publish failure report, to wit:
```
  "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1535.7-1535.130: error: The element “dfn” must not appear as a descendant of the “dfn” element.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1535.131-1535.285: error: Start tag “a” seen but an element of the same type was already open.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1535.131-1535.285: error: End tag “a” violates nesting rules.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1536.7-1536.103: error: The element “dfn” must not appear as a descendant of the “dfn” element.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1536.104-1536.233: error: Start tag “a” seen but an element of the same type was already open.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1536.104-1536.233: error: End tag “a” violates nesting rules.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1536.269-1536.272: error: Stray end tag “a”.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1536.269-1536.278: error: Stray end tag “dfn”.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1536.269-1536.282: error: Stray end tag “a”.
    "file:/home/runner/work/rdf-semantics/rdf-semantics.w3c/spec/index.html":1536.269-1536.288: error: Stray end tag “dfn”.
```

Now, those errors appear to be about lines 1535 and 1536 — 
```

  <p>Both of these can be handled by allowing the rules to apply to a generalization of the RDF syntax
```
— which clearly do not contain the tags flagged by Echidna.

_Hopefully,_ once my change request is merged, which will re-trigger Echidna and PR-Preview, the issues will magically resolve, _or_ the errors will be revised and point to lines that actually contain problematic tags.

pfps:
These may all be the result of malformed XML around line 202.  I'll put in a PR to try to fix the problem.

pfps:
I'm leaving the needs discussion label, not because there hasn't been discussion but because the final version of the PR needs careful consideration.

domel:
According to telecon and different suggestions, I've just changed CSS for the tables here.  If you accept this layout for small devices, I will implement these changes in other specs.

gkellogg (approved):
Seems pretty readable across a variety of devices.

domel:
I have prepared other PRs (some of them I improved more if needed). So I'm waiting for 'yes'.

pfps:
I thought the changes were going to be conditioned on the type of device.  Why is this not happening?

It appears that the width of tables are being restricted to 80% of something.  This seems wrong to me.  Why not allow tables to be as wide as that thing?

This document is not a very good one for testing out the changes as there are only a few tables with more than one column and none of them are very wide.

TallTed:
Looks far better to me, overall. I see breaks are not happening mid-word. This does mean that some side-scrolling will be necessary in some smaller-windowed environments. Echidna errors appear to have been resolved.

@pfps — I'm guessing that "It appears that the width of tables are being restricted to 80% of something" because you see `table { word-break: normal; overflow-wrap: anywhere; max-width: 80vw; }`, which does indeed limit the table width to 80% of the viewport, which is the browser window. However, if you have large viewports available, you should see that this limit is subordinate to other limitations, so doesn't actually come into play. If you have small viewports available, you should see that this limit doesn't get enforced because the unbroken words cause the table cells to be wider than this setting, to contain those unbroken words.

domel:
@pfps I get the impression that you often try to question things, but you don't always give specifics. If you question my HTML and CSS skills, please indicate what to improve and block the possibility of merge. Otherwise, accept.
PR is prepared in accordance with W3C recommendations for RWD and CSS media queries. It makes no sense to indicate specific (types of) devices because there are thousands of them. What's worse, in a year or two everything can change. It is good practice to check the size of the display.

pfps:
As far as I can tell, the 80vw is limiting the width of tables in some places and, again as far as I can tell, the limit is active.  For example, in the PR preview for some window widths the RDFS Semantic Conditions table is narrower than the text width and some lines overflow when at the same width they do not overflow without the PR.

So, again, why limit table width?

domel:
Because 80vw is the default width in respec for the content. Note that 10vw are left and right margin.

pfps:
This does not match with what I see.  With the PR the width of the RDFS Semantic Conditions table is narrower than without the PR.  So if there is a change, how can 80vw be the default?

In any case, if 80vw is the default then there should be no reason to include it (as it would currently be doing nothing) and a good reason to not include it (as it would override future changes in RESPEC).

TallTed:
> if 80vw is the default then there should be no reason to include it

I wonder if it wouldn't be worth creating a (or a few) separate PR for testing, perhaps made against one of our documents with more tables, especially tables that are problematic before whatever CSS tweak we find to be best?

TallTed:
@pfps — If you could provide links (including fragment ID) to the table(s) you're seeing as problematic, that will help me (and, I imagine, others) to dig into possible solutions.

pfps:
As stated above, it's the RDFS semantic conditions table.  As far as I can tell there is no direct link to it.  I've always either scrolled through the document or searched for the caption.

TallTed:
@pfps — Do you mean the table immediately above [this anchor](https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/30.html#RDFS_axiomatic_triples)? (which should indeed have its own ID! as I've added in [PR#33](https://pr-preview.s3.amazonaws.com/TallTed/rdf-semantics/pull/33.html#rdfs_semantic_conditions)), that has only one column?

Or the table following [this anchor](https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/30.html#rdfs_patterns), now also at [this direct link from PR#33](https://pr-preview.s3.amazonaws.com/TallTed/rdf-semantics/pull/33.html#rdfs_entailment_patterns) that has 3 columns of odd widths (left and right columns have odd wraps, and center could lose a bit of width to eliminate those wraps)...?

pfps:
There is only one table for RDFS semantic conditions - the one with RDFS semantic conditions as its caption - the first one.  The other is RDFS entailment patterns.

pfps:
The other table is actually damaged in the PR when I look at it.   In some of the LH columns the rule name is split, e.g., rdfs4 is on one line and a is on the next.   This is not the case without the PR.

I thought that there was not going to be any word splitting going on.

gkellogg:
I’ve noted similar issues in Turtle and TriG, where tables view properly until the width is reduced enough, and then they shrink to narrower than surrounding content.

domel:
Could you write more details? In what size of screen did you see that effect?

gkellogg:
I simply narrowed the window I was looking at in Safari on my Mac, when it got down to around t 700-900 pixels wide, the table became narrower than the surrounding text. Before, it was the same width.

domel:
OK, I will fix that. But remember that the overwhelming majority of mobile devices are covered by my patch. I tested it on 46 mobile devices. But ok, I'll try to work it out a bit more ( I mean in all PRs about RWD).

pfps:
It's not testing on mobile devices that is required, but testing on non-mobile devices.   

A requirement for this PR is that it have no negative effects on non-mobile devices.

pfps (changes requested):
This PR needs to be tested out on all documents to certify that it does not have negative effects on any of them when viewed on non-mobile devices.

TallTed:
Welcome to HTML+CSS hell. Things that shouldn't happen, do. Things that should happen, don't.

The table with the "RDFS semantic conditions" caption has only one column, so it's odd that it's a table at all. That said, I see nothing problematic with it. The `80vw` setting is a *maximum* width, not an *always* width, so it being slightly smaller than that max (as I see in all viewports that are sufficiently large), which is wide enough that no words split.

In the other table, the unfortunate break/wraps of the first column occur when the value (here, `rdfs`) is wrapped by spaces, as —
```html
<dfn id="dfn-rdfs1" tabindex="0" aria-haspopup="dialog" data-dfn-type="dfn"> rdfs1 </dfn>
```
— but do *not* occur when the value is *not* wrapped by spaces, as —
```html
<dfn id="dfn-rdfs1" tabindex="0" aria-haspopup="dialog" data-dfn-type="dfn">rdfs1</dfn>
```

(I'm only viewing this is current Chrome, at the moment. If there's a known problematic browser, I can try viewing and analyzing there, too.)

gkellogg:
I still don't understand why the table in 9.2.1 Patterns of RDFS entailment is the full-width of the text area until the window is shrunk sufficiently, then it is narrower than the text width. It's better than before, and perhaps good enough.

domel:
@gkellogg If I understand you correctly, it depends on the content. The table takes up less than the entire webpage (on small screens) because all lines fit and it does not need to expand anymore.

pfps:
Have all the deleterious effect of previous versions of the PR been eliminated?

domel:
@pfps Yes, it seems to me that all deleterious effects (although I wouldn't call it that) have been corrected. This applies to all other PRs, which differ from each other because their structure is different, but the effect is the same.

pfps (changes requested):
I see even worse behaviour now.

Not only does the RDFS Semantic Conditions table (the yellow table in https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/30.html#rdfs_interpretations) shrink horizontally more than regular lines, but now the font shrinks.

Please completely vet the changes  before asking for yet another review.

domel:
@pfps 

> I see even worse behaviour now.

It's your subjective feeling.


> Not only does the RDFS Semantic Conditions table (the yellow table in https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/30.html#rdfs_interpretations) shrink horizontally more than regular lines, but now the font shrinks.

Yes exactly. This is how it solves the issue of small screens, small screens need a smaller font because they have a smaller surface area. By the way, this effect has been around for a long time.

> Please completely vet the changes before asking for yet another review.

I checked the changes carefully. If you don't like them, I invite you to change the HTML and CSS. I will not change anything because I believe that this is the best possible solution, many times better than the original (before PR).

pfps:
No, these are objective findings. 

Having table width constrained to be less than line width and having that cause the table to consume more vertical space is an objective deleterious effect.

Having the font size of a table get smaller when there is still room on the right-hand margin is an objective deleterious effect.

domel:
As I said above, I disagree. Note that each table has a different width, so the effect you're imagining would require you to write separate styles for each table. And in the future, anything you add to these tables will likely make the table too small to fit on smaller screens again. In short, I do not want and will not change anything. As I understand it, you can continue working on your own, or go back to the original version, which renders much worse on small screens.
BTW, those yellow areas shouldn't be tables at all, that's against the sematics of HTML elements.

pfps:
No.   There is no reason whatsoever to leave blank space on the right-hand side of tables when regular lines are longer and the table font is reduced or lines are unnecessarily split.  This is not a matter of the natural width of tables - many tables have their width unnecessarily reduced and this reduction has deleterious effects.

domel:
Okay, so I'm waiting for your code and let me comment on it.

pfps:
I am not the one that wants changes.  I view the current situation as much better than any of the PRs that have been proposed.  I am happy leaving things as they are.

domel:
If you think that on mobile phones the original version displays better than the current one, then I can't talk to you. I don't think there's any point in wasting any more of your time or mine. However, I advise you to take a few mobile phones, check and rethink the situation. One more question I would ask myself if I were you is which specific devices (and what percentage of all) show the defect you are talking about (because manually resizing the desktop browser is not a determinant here).

pfps:
My investigations have only been on my laptop.   I have never implied anything about behaviour on mobile phones.  All I have ever stated is that every version of the PR is significantly worse viewed on my laptop than the status without the PR.  

My view is that the PR is unacceptable if it makes things worse when viewing the document on a laptop and for every version of the PR this has been the case.

pfps:
Here are two screenshots, one with PR 30 and one without.   In the one with, the width of the table has been unnecessarily reduced and the font size also unnecessarily reduced.  This only happens in some window widths.


Screenshot from 2023-05-16 18-41-27
Screenshot from 2023-05-16 18-42-06

Many other tables have similar problems.

gkellogg (commented):
Changes to other specs seem to have solved the problems related to the table narrowing below the line length when shrinking the window, but I still see it here. I'm fine with the text size shrinking in these cases. None of these changes should affect normal behavior with reasonably sized windows on laptops, but it would be nice to get this sorted out. It doesn't make sense to loose substantial horizontal space when it is more needed.

pfps:
@gkellogg I run with small fonts (relative to screen distance) so further shrinkage is bad for me.   At some point it may be better to shrink fonts but the breakpoint is much too high for me with this PR.

This is the first subjective issue that I have reported.

domel:
@pfps You can now see if you like it on your super small computer monitor.

pfps:
My laptop is a Yoga C940 with a 14" screen.  I suppose that this may seem super small compared to the gigantic phones and phablets that are in vogue now, but I find it just the right size for me.

In any case, as far as I can tell it is not the size of the screen that matters, but how wide the browser window is.   It is quite common to have multiple windows on screens with them dividing up the the horizontal space somehow.   This makes it routine for the width of a window to be quite small even on large screens.

domel:
@pfps OK, nevermind. Let's get to the specifics. Are you satisfied with the current commit?

gkellogg (approved):
LGTM.

pfps:
Here are two screenshots that show an objective degredation with the current PR.   When the window reaches a certain width the font size in table is immediately reduced with no reference to any benefit there might be to overcome this degredation.  For both tables in these screenshots there is zero benefit.

Screenshot from 2023-05-17 18-13-53
Screenshot from 2023-05-17 18-15-27

For me, this changes the table from being easily readable to not being easily readable.

pfps:
A separate question is whether this kind of font hackery is ever justified.   I'm not an expert in user interface design to determine this, but some expert should probably be consulted.

domel:
> For me, this changes the table from being easily readable to not being easily readable.

It's depend on how you define "readable". It's definitely more readable in mobile phones and tablets because of no need to scrolling. It's the same readable in normal laptops and desctop (the font size is the same). We probably won't please everyone, but the solution currently covers as many of the cases as possible. As I wrote earlier, I have been professionally involved in front-end development for a long time and if you feel the need to check with another expert, I have no fear that you will hear the same.

domel:
Back to the font sizes themselves.  I'm shocked that you don't mind the font size in the pink and green areas and in the yellow areas the same font size is a problem for you.

TallTed:
It's hard to see the difference between those screenshots, with them stacked vertically. This may be helpful to others:

| wider, has bigger font in the tables | narrower, has smaller font in the tables |
|:-----:|:--------:|
|  |  |

I agree with @pfps that this makes the tabular data harder to read, even at the substantially blown-up size seen in these screenshots, vs the font sizes seen in the actual document.

domel:
@TallTed ok, i propose to accept the PR (which is about CSS) and open a new one to refractor HTML, yellow, pink etc areas are the tables and they shouldn't (in HTML table element is for tabular data and it is not a tabular data). So I propose to change HTML tables there to block element (that should looks like the normal text including font size).

TallTed:
@domel — I agree that the pictured tables shouldn't be tables. Replacing these with block elements seems reasonable, presuming the background and text color effects are preserved. 

That said, I think that the font size changes observed by @pfps are likely CSS-related, so I think @pfps should look at the *actually* tabular data in the other tables, and see whether they do the same font-size-shrinking with the same action — as if they do, this issue #30 should remain open and the CSS driving that should be addressed.

domel:
All 10 PRs follow the same font size rules as this PR.  Of course, if you want, I can change it, but it will be an exception.  The other specs (that you have accepted) do just that. So what is the final decision?

TallTed:
@domel — Have you observed the font size changes reported by @pfps when changing windows sizes? They're a problem, wherever they may occur. I am surprised that they only occur (or have only been reported) in one document and by one user — so maybe we should check that by getting browser and OS details, which may be having an impact. Maybe we should have other people check with that browser on their own OS, and have that person check other browser(s) on their OS.

Please understand that I'm not deciding nor dictating anything — only confirming that an issue reported by another viewer appears to be an actual issue, which should be resolved by *some* means, and CSS seems the right place for that fix. Elsewhere, I've reported my observations that the CSS changes there *appear* to have had the desired effect, but other viewers may report differently, and their observations should not be treated as any less important than mine.

domel:
@TallTed I can't agree that this is a problem, on the contrary, thanks to this, the text has a chance to fit on small screens.  You can go into the style editor in your browser (probably by pressing F12) and check that increasing the font size destroys responsiveness (in all spec with tables). In short, I can't do it the way you want it.

RWD is not about making the website look identical on a small smartphone and a huge monitor on 2 desktops, it should look so that it is legible.

TallTed:
Again, I did not say "increase the font size".

I (and @pfps) said "When the viewport/window reaches some width smaller than full-screen, the font size of the content of these tables gets smaller, while the font size of the surrounding body remains as it was."

That remains a problem.

domel:
No, it's normal behavior if you have media queries. It's the main feature of media queries that when it grows to its size, it changes styles.  In this case, among other things, font size.  You will find it on all websites in the world.

TallTed:
> You will find it on all websites in the world.

This is trivially provable as inaccurate. At best, it is accurate only for those web pages which use "media queries" which is far from "all websites in the world."

Further, your response to @pfps said that in [his screenshots](https://github.com/w3c/rdf-semantics/pull/30#issuecomment-1552309131), the font sizes were identical throughout, and you couldn't understand why he saw otherwise. Now you're saying, the smaller font size in the tables is the desired and intended effect. @pfps and I are both saying, that is an undesirable and problematic effect!

domel:
From the beginning I say that the font is smaller by I reduced it myself.  Thanks to this, the spec is finally responsive and I can view it on my phone right now.

domel:
However, I do not understand this fuss about fonts a bit, since you agree with me that the yellow areas should be rewritten in HTML (as a block elements) and then they will meet the assumptions of RWD and will have a large font.  Everything you want.

TallTed:
I guess we should <strike>table this</strike> set this discussion aside for now, and see how the remaining tables look after the shouldn't-have-been-tables are made into blocks.

pfps:
From a recent version of the PR:

```
 @media (max-width: 850px) {
        table th, table td { font-size: 12px; padding: 3px 4px;}
        .othertable {padding: 4px;}
        img {max-width: 85vw;}
    }  
   ```

So, although I am not a CSS expert by any means, it appears to me that the font size of all tables was forcibly changed to 12pc from whatever it would otherwise be when the window width was under a certain value.  In all cases that I managed to test, this resulted in smaller fonts in tables.

 Why did no-one else notice?   Probably because they didn't test this PR where it actually changed the display of the document.

**Comment by domel** [comment]
Not true. Here is the recent version
```
    @media (max-width: 550px) {
        table th, table td { font-size: 14px; padding: 3px 4px;}
        table td.semantictable, table td.ruletable { font-size: 14px;}
        .othertable {padding: 4px;}
        img {max-width: 85vw;}
    }  
```

pfps:
See https://github.com/w3c/rdf-star-wg/issues/61

ktk:
@pchampin will talk to the accessibility team at W3C to get some input on the issue.

TallTed (changes requested):
I would like the dynamic font size change to be removed. I may request other changes if that resizing is determined to be appropriate by `a11y`/accessibility folks.

pchampin:
Trying to move forward with this...

1. I believe that *before* making this document mobile-friendly, we should ensure that we are happy with the base HTML/CSS, which I believe is not the case (e.g. many 1-column "tables" should probably not use `<table>`).
2. Then, we need to determine what is the best way to make the documents mobile friendly
  To that effect, I reached out to W3C CSS/a11y people, still waiting for their response.

In the meantime, I think the problem of this PR is that it tweaks too many parameters at the same time (wrap, font-size, width). My strategy would be to change one parameter at a time, and evaluate at each step the pros and cons.

domel:
@pfps As I wrote few weeks ago, now there are some places in the spec that use table elements but semanticly they are not a tables. They are blocks so it should uses block elements e.g. section elements. Thanks @pchampin for writing this obvious truth.

pchampin:
> @pchampin I don't understand what is wrong with one-column tables. I also don't know what would replace one-column tables.

* All axiom "boxes" (on lavender background) contain a *single cell*. They hardly qualify as "tables".
* The "boxes" containing semantic conditions (on yellow background) are more akin to lists than to tables, MO.

I agree that the difference can be blurry and somewhat subjective, but a pedantic reading of [the HTML spec](https://html.spec.whatwg.org/multipage/tables.html#the-table-element) excludes the use-cases above:

> The [table](https://html.spec.whatwg.org/multipage/tables.html#the-table-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) data with **more than one dimension**, in the form of a [table](https://html.spec.whatwg.org/multipage/tables.html#concept-table).

(emphasis is mine)

pfps:
I'm a bit puzzled how tables can represent data with, say, 1000 dimensions.  So I don't give much weight to that statement.

As far as I can tell, tables are good for representing sequences in a way that emphasizes the commonality of the elements of the sequence.   However, if there is a way of showing a sequence that puts the elements in equal-width boxes with no extra space between them, I'm not adverse to changing the HTML (even though it will be a pain for essentially no gain).

There are several tables that have multiple rows and multiple columns so it seems that there is content that is ideally suited be tables.

domel:
> However, if there is a way of showing a sequence that puts the elements in equal-width boxes with no extra space between them, I'm not adverse to changing the HTML

Yes, it is very easy to replace with any block element.

pfps:
Can someone provide me an example of how a block can be used to substitute for one-element and one-dimensional tables?

pchampin:
I discussed this issue with @bert-github, and my take-home messages are the following:
- we should indeed try to avoid horizontal scrolling on small screens
- reducing the font size and changing word wrap are however last resort solutions
- another option to make tables mobile friendly is to change their layout on small screens; I made a proposal in that sense #40 
- the lists of axioms ([example](https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/40.html#RDFS_axiomatic_triples)) and of semantic conditions ([example](https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/40.html#rdfssemcond1)) are not semantically tables, and should rather use other tags (typically `<ul>`). Note that this could be done while keeping the exact same presentation.

pfps:
The remaining problem is that changes were made to the other documents, contrary to what was supposed to happen.   These changes should be undone.

PR #45, opened 2023-09-22
add entailment rule for datatypes

### LINKED ISSUE #44: small bug in axiomatization of RDF inference
It appears that a needed rule of inference was left out of the set of RDF inference rules.  I expect that what happened is that when datatypes were added to RDF inference the inference rules from datatype inference were just copied over and the interaction between datatypes and rdf:type was not considered.

The rule that I think need to be added is:

rdfD1a	for ddd in D with non-empty value space	 _:nnn rdf:type ddd .

which comes from the second RDF semantic condition.

Then places where the RDF rules are used need to include this rule.

pfps:
Fixes #44

The basic issue is that RDF entailment recognizes the rdf:langString and the xsd:string datatypes.  This means that the domain of discourse includes all XSD strings and all language-tagged strings.   In these interpretations the following triples are true:
```
_:b rdf:type rdf:langString .
_:c rdf:type xsd:string.
```
and thus need to be accounted for in the entailment patterns.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/45.html" title="Last updated on Nov 2, 2023, 4:02 PM UTC (290ffda)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/45/295f6fa...290ffda.html" title="Last updated on Nov 2, 2023, 4:02 PM UTC (290ffda)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2061,7 +2061,7 @@
   <h2>Substantive changes since RDF 1.1</h2>
 
   <ul>
-    <li> The RDF entailment rule <a>rdfD1a</a> was added.  This rule was not added for RDF entailment when the two built-in dataypes were added to RDF entailment. </li>
+    <li> RDF entailment rule <a>rdfD1a</a> was added in RDF 1.2.  This rule should have been included in RDF 1.1 when the two built-in datatypes (<code>xsd:string</code> and <code>rdf:langString</code>) were added to RDF entailment. </li>
   </ul>
 </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -1033,6 +1033,25 @@
       <p><code>ex:a ex:p _:x . <br/>
        _:x rdf:type xsd:integer . </code></p>
 
+      <p>The last semantic condition above also justifies the following entailment pattern:</p>
+
+      <table>
+        <tbody>
+          <tr>
+            <th > </th>
+            <th ><strong>any S</strong></th>
+            <th ><strong>then S RDF entails, recognizing D</strong></th>
+          </tr>
+          <tr >
+            <td class="othertable"><dfn>rdfD1a</dfn></td>
+            <td class="othertable">for ddd in D with non-empty value space</td>
+            <td class="othertable">_:nnn <code>rdf:type</code> ddd <code>.</code></td>
+          </tr>
+        </tbody>
+      </table>
+
+
+
       <p>In addition, the first RDF semantic condition justifies the following entailment pattern:</p>
 
       <table>
@@ -1588,7 +1607,7 @@
     <li>Add to S all the RDF (and RDFS) axiomatic triples which do not contain any container membership property IRI.</li>
     <li>For each container membership property IRI which occurs in E, add the RDF (and RDFS) axiomatic triples which contain that IRI.</li>
     <li>If no triples were added in step 2., add the RDF (and RDFS) axiomatic triples which contain <code>rdf:_1</code>.</li>
-    <li>Apply the rules <a>GrdfD1</a> and <a>rdfD2</a> (and the rules <a>rdfs1</a> through <a>rdfs13</a>),
+    <li>Apply the rules <a>GrdfD1</a>, <a>rdfD1a</a>, and <a>rdfD2</a> (and the rules <a>rdfs1</a> through <a>rdfs13</a>),
       with D={<code>rdf:langString</code>, <code>xsd:string</code>}, to the set in all possible ways, to exhaustion.</li>
   </ol>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -2061,7 +2061,7 @@
   <h2>Substantive changes since RDF 1.1</h2>
 
   <ul>
-    <li> None so far. </li>
+    <li> The RDF entailment rule <a>rdfD1a</a> was added.  This rule was not added for RDF entailment when the two built-in dataypes were added to RDF entailment. </li>
   </ul>
 </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -1038,11 +1038,11 @@
       <table>
         <tbody>
           <tr>
-            <th > </th>
-            <th ><strong>any S</strong></th>
-            <th ><strong>then S RDF entails, recognizing D</strong></th>
+            <th> </th>
+            <th><strong>any S</strong></th>
+            <th><strong>then S RDF entails, recognizing D</strong></th>
           </tr>
-          <tr >
+          <tr>
             <td class="othertable"><dfn>rdfD1a</dfn></td>
             <td class="othertable">for ddd in D with non-empty value space</td>
             <td class="othertable">_:nnn <code>rdf:type</code> ddd <code>.</code></td>


spec/index.html
```diff
@@ -1033,6 +1033,25 @@ <h4>Patterns of RDF entailment (Informative)</h4>
...
+      <p>The last semantic condition above also justifies the following entailment pattern:</p>
+
+      <table>
+        <tbody>
+          <tr>
+            <th > </th>
+            <th ><strong>any S</strong></th>
+            <th ><strong>then S RDF entails, recognizing D</strong></th>
+          </tr>
+          <tr >
```
TallTed, line 1045: Just dropping some unnecessary space chars
```suggestion
            <th> </th>
            <th><strong>any S</strong></th>
            <th><strong>then S RDF entails, recognizing D</strong></th>
          </tr>
          <tr>
```

TallTed (changes requested):
Is line 1610 (previously 1591) the only "[place] where the RDF rules are used"? If not, #44 should be kept open until all those places are also addressed.

gkellogg:
PR Preview is enabled for this repo, and is on previous PRs. It’s not perfect, though, and it doesn’t seem to have completed in this case.

pfps:
> Is line 1610 (previously 1591) the only "[place] where the RDF rules are used"? If not, #44 should be kept open until all those places are also addressed.

I believe that the only place the rules are used is in Appendix A, with the reference to them on the new line 1610.

gkellogg:
Does it need some new tests?

pfps:
I think so.  

A test would be the empty graph entailing 
_:y rdf:type xsd:string .
_:z rdf:type rdf:langString .
in RDF entailment.
If there is a new string datatype that is required for RDF entailment then that should be included as well

If there are tests for D-entailment then the empty graph entailing
_:x rdf:type xsd:integer .
in D-entailment recognizing xsd:integer should be included.

gkellogg:
So as not to loose this maybe best to create an issue referencing https://github.com/w3c/rdf-semantics/pull/45#issuecomment-1738025041 in https://github.com/w3c/rdf-tests, or keep this issue open with a "needs tests" label.

gkellogg:
Also, this will need an entry in "Changes since ...".

pfps:
Added entry to changes since RDF 1.1.

pfps:
How is the WG going to handle changes that need tests?   Is the PR going to remain unmerged until the tests are done?

TallTed:
> How is the WG going to handle changes that need tests? Is the PR going to remain unmerged until the tests are done?

I suggest creation of an issue ("need test for PR xxx") such that accepted PRs may be merged more quickly, allowing other work to proceed, while a test PR is pending.

pfps:
OK, I'll add a test issue when (if?) this PR is merged.   As this and other PRs that need tests should have a needs test tag it should be possible to check that each of these PRs have a resolved test issue.

pchampin:
Alternatively, we can put the needs-test on #44 and keep it open until the test is created?

spec/index.html
```diff
@@ -2042,7 +2061,7 @@ <h2>Acknowledgments</h2>
   <h2>Substantive changes since RDF 1.1</h2>
 
   <ul>
-    <li> None so far. </li>
+    <li> The RDF entailment rule <a>rdfD1a</a> was added.  This rule was not added for RDF entailment when the two built-in dataypes were added to RDF entailment. </li>
```
pchampin, line 2064: I find these sentences confusing. Both use the past tense, but I'm assuming that they are referring to different points in the past. More precisely, I'm assuming (without being certain) that they mean: "The [rule] was added *(in 2023)*. This rule was not added *(in 2014)*...". In any casen this needs clarification.

I would suggest replacing, in the 2nd sentence, "was not added" by "had not been added", but I defer to native speakers to find the best wording.

pchampin (approved):
Except for the sentence in the change log that needs some clarification, LGTM.

gkellogg:
> Alternatively, we can put the needs-test on #44 and keep it open until the test is created?

Creating a test in rdf-tests that references #44 should be adequate; after all, it's a testing issue not a spec/editorial issue.

spec/index.html
Antoine-Zimmermann: The two built-in datatypes of RDF 1.1's RDF entailment are `xsd:string` and `rdf:langString`.
```suggestion
    <li> The RDF entailment rule <a>rdfD1a</a> was added in RDF 1.2.  This rule should have been included in RDF 1.1 when the two built-in datatypes (<code>xsd:string</code> and <code>rdf:langString</code>) were added to RDF entailment. </li>
```

Antoine-Zimmermann:
Like @pchampin, I approve these changes pending the change to the sentence in the change log.

However, I noticed that the algorithm in the Appendix is still incomplete with this rule. I opened issue #46 to discuss this.

doerthe:
The entailment only holds, if the set of literals is fixed (and contains at least one xsd:string and one rdf:langString), I guess that is the case from our point of view?

doerthe:
I don't like the rule (yet?) and would argue that a model normally only works with the iris and literals occurring in the graphs plus a fixed (finite!) set of iris which are a fixed part of the language.

But I am undecided, so please convince me here :)

pfps:
> This would thus mean that, if we'd ever allow literals in subject position we would have to add a triple for all possible xsd-literals we can ever construct?

Yes.  It is a consequence of datatype semantics that if l is in the lexical space of d then l rdf:type d . is true in all interpretations recognising d.  So the rule GrdfD1 is not complete for generalized RDF.

pfps:
> I don't like the rule (yet?) and would argue that a model normally only works with the iris and literals occurring in the graphs plus a fixed (finite!) set of iris which are a fixed part of the language.
> 
> But I am undecided, so please convince me here :)

In the original RDF semantics interpretations were with respect to a vocabulary.  For entailment to work correctly the vocabulary had to include all IRIs in both graphs in the entailment.  That is, the empty graph did not RDFS-entail :a rdf:type rdfs:Resource .  which is weird. 

The RDF 1.1 simple semantics requires that all IRIs have a denotation in IR union IP.   Then RDF 1.1 interpretations require that all IRIs have a denotation in IR because of the definition of rdf:property.  So in the RDFS 1.1 semantics every IRI has a denotation that is in ICEXT(rdf:Resource).

doerthe:
I am thinking about the consequence that the domain of discourse could thus never be finite. 

I also wonder whether I could somehow violate the compactness theorem here if I had negation and literals in subject position, but that did not work for my quick thought experiment, because I cannot do something like _:x^^xsd:string, that is, I can still not quantify over my strings...

pfps:
The domain of discourse is not finite in RDF intepretations because of the built-in datatypes.   However, there is no requirement that the denotation of different IRIs is different so the non-literal part of the domain of discourse can be finite.

spec/index.html
TallTed: Just one tweak to AZ's update, to drop the initial `The`
```suggestion
    <li> RDF entailment rule <a>rdfD1a</a> was added in RDF 1.2.  This rule should have been included in RDF 1.1 when the two built-in datatypes (<code>xsd:string</code> and <code>rdf:langString</code>) were added to RDF entailment. </li>
```

doerthe:
OK, I thank Peter for saving us from having the discussions from RDF 1.1 again. I understand why we do not want to depend on the vocabulary  in this case and also approve.

doerthe (approved):

PR #50, opened 2024-10-31
add in RDF dataset definitions from SPARQL query

pfps:
Addresses https://github.com/w3c/sparql-query/issues/155

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1498,6 +1498,60 @@
     between datasets would require RDF graph entailments between the graphs with the same name
     (adding in empty graphs as required).</p>
 
+  <table>
+    <caption>Definition of an RDF dataset</caption>
+    <tr>
+       <td class="semantictable">
+       <p>An <dfn data-lt="RDF dataset">RDF dataset</dfn> I is a set:</p>
+              { G, (&lt;u<sub>1</sub>&gt;, G<sub>1</sub>), (&lt;u<sub>2</sub>&gt;, G<sub>2</sub>), . .
+              . (&lt;u<sub>n</sub>&gt;, G<sub>n</sub>) }<br>
+              where n&ge;0 and G and each G<sub>i</sub> are graphs, and each &lt;u<sub>i</sub>&gt; is an IRI. Each
+              &lt;u<sub>i</sub>&gt; is distinct.</p>
+            <p>G is called the default graph. (&lt;u<sub>i</sub>&gt;, G<sub>i</sub>) are called named
+              graphs.</p>
+        </td>
+    </tr>
+  </table>
+  <table>
+    <caption>Definition of active graph in an RDF dataset</caption>
+    <tr>
+       <td class="semantictable">
+            <p>The <dfn data-lt="active graph">active graph</dfn> in an RDF dataset is the graph from the dataset used for basic graph pattern
+              matching.</p>
+       </td>
+    </tr>
+  </table>
+
+  <table>
+    <caption>Definition of RDF dataset merge</caption>
+    <tr>
+       <td class="semantictable">
+            <p>Let DS1 = { G1, (&lt;u1<sub>1</sub>&gt;, G1<sub>1</sub>), (&lt;u1<sub>2</sub>&gt;,
+              G1<sub>2</sub>), . . . (&lt;u1<sub>n</sub>&gt;, G1<sub>n</sub>) },<br>
+              and DS2 = { G2, (&lt;u2<sub>1</sub>&gt;, G2<sub>1</sub>), (&lt;u2<sub>2</sub>&gt;,
+              G2<sub>2</sub>), . . . (&lt;u2<sub>m</sub>&gt;, G2<sub>m</sub>) }</p>
+            <p>The <dfn data-lt="RDF dataset">RDF dataset merge</dfn> of DS1 and DS2 is:<br>
+              DS={ G, (&lt;u<sub>1</sub>&gt;, G<sub>1</sub>), (&lt;u<sub>2</sub>&gt;, G<sub>2</sub>), .
+              . . (&lt;u<sub>k</sub>&gt;, G<sub>k</sub>) }<br>
+              where:</p>
+            <p>Write N1 for { &lt;u1<sub>j</sub>&gt; j = 1 to n }<br>
+              Write N2 for { &lt;u2<sub>j</sub>&gt; j = 1 to m }<br></p>
+            <ul>
+              <li>G is the <a data-cite="RDF12-SEMANTICS#dfn-merge">merge</a> of G1 and G2
+              </li>
+              <li>(&lt;u<sub>i</sub>&gt;, G<sub>i</sub>) where &lt;u<sub>i</sub>&gt; is in N1 but not
+                in N2</li>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1498,6 +1498,60 @@ <h2>RDF Datasets</h2>
...
+  </table>
+  <table>
+    <caption>Definition of active graph in an RDF dataset</caption>
+    <tr>
+       <td class="semantictable">
+            <p>The <dfn data-lt="active graph">active graph</dfn> in an RDF dataset is the graph from the dataset used for basic graph pattern
+              matching.</p>
+       </td>
+    </tr>
+  </table>
```
afs, line 1523: This is SPARQL specific and can be left in SPARQL Query. It changes during query execution (GRAPH).

pfps: Done

domel:
Why definitions are in tables? Such text certainly has nothing to do with the table.

pfps:
> Why definitions are in tables? Such text certainly has nothing to do with the table.

Just because all the definitions were done that way by Pat for 1.1.

afs (approved):

spec/index.html
```diff
@@ -1498,6 +1498,51 @@ <h2>RDF Datasets</h2>
     between datasets would require RDF graph entailments between the graphs with the same name
     (adding in empty graphs as required).</p>
 
+  <table>
+    <caption>Definition of an RDF dataset</caption>
+    <tr>
+       <td class="semantictable">
+       <p>An <dfn data-lt="RDF dataset">RDF dataset</dfn> I is a set:</p>
```
gkellogg, line 1505: RDF Dataset is actually defined in RDF Concepts, although it has an alias here, so it's sufficient to just use an anchor.
```suggestion
       <p>An <a>RDF dataset</a> I is a set:</p>
```suggestion

File: spec/index.html [hunk 3]
```diff
@@ -1498,6 +1498,51 @@ <h2>RDF Datasets</h2>
...
+
+  <table>
+    <caption>Definition of RDF dataset merge</caption>
+    <tr>
+       <td class="semantictable">
+            <p>Let DS1 = { G1, (&lt;u1<sub>1</sub>&gt;, G1<sub>1</sub>), (&lt;u1<sub>2</sub>&gt;,
+              G1<sub>2</sub>), . . . (&lt;u1<sub>n</sub>&gt;, G1<sub>n</sub>) },<br>
+              and DS2 = { G2, (&lt;u2<sub>1</sub>&gt;, G2<sub>1</sub>), (&lt;u2<sub>2</sub>&gt;,
+              G2<sub>2</sub>), . . . (&lt;u2<sub>m</sub>&gt;, G2<sub>m</sub>) }</p>
+            <p>The <dfn data-lt="RDF dataset">RDF dataset merge</dfn> of DS1 and DS2 is:<br>
```
gkellogg, line 1524: This actually becomes a duplicate definition of RDF dataset, whereas I believe you want a definition of "RDF Dataset merge".

```suggestion
            <p>The <dfn>RDF dataset merge</dfn> of DS1 and DS2 is:<br>
```

gkellogg (changes requested):
RDF dataset is defined in RDF Concepts, not Semantics; it is RDF dataset merge which should be defined here.

Note that some of the markup is wrong (`data-lt` overrides the text content of the element, which is not always what you want).

afs:
> RDF dataset is defined in RDF Concepts, not Semantics; it is RDF dataset merge which should be defined here.

We need notation for "RDF dataset" for use in SPARQL, in the definitions of "RDF Dataset Merge".
RDF Concepts does not provide that notation. 

If it were in  RDF concepts, quoting here (without the heading "Definition of an RDF dataset") would help the reader.

The section RDF Concepts section [RDF Dataset Comparison](https://www.w3.org/TR/rdf12-concepts/#section-dataset-isomorphism) would also benefit (as it is currently, it can be read as only applying to datasets with one named graph).

spec/index.html
pfps: Should be fixed in latest commit.

spec/index.html
pfps: The dfn markup is copied from examples elsewhere.  If they are wrong here then I think there are lots that have to be changed throughout the document set.

spec/index.html
```diff
@@ -1498,6 +1498,52 @@ <h2>RDF Datasets</h2>
...
+
+  <table>
+    <caption>Definition of RDF dataset merge</caption>
+    <tr>
+       <td class="semantictable">
+            <p>Let DS1 = { G1, (&lt;u1<sub>1</sub>&gt;, G1<sub>1</sub>), (&lt;u1<sub>2</sub>&gt;,
+              G1<sub>2</sub>), . . . (&lt;u1<sub>n</sub>&gt;, G1<sub>n</sub>) },<br>
+              and DS2 = { G2, (&lt;u2<sub>1</sub>&gt;, G2<sub>1</sub>), (&lt;u2<sub>2</sub>&gt;,
+              G2<sub>2</sub>), . . . (&lt;u2<sub>m</sub>&gt;, G2<sub>m</sub>) }</p>
+            <p>The <dfn data-lt="RDF dataset merge">RDF dataset merge</dfn> of DS1 and DS2 is:<br>
```
gkellogg, line 1525: It is redundant to use `data-lt` with the value which is the same as the element content.

```suggestion
            <p>The <dfn>RDF dataset merge</dfn> of DS1 and DS2 is:<br>
```

gkellogg (approved):

gkellogg (approved):
I pushed some changes for "merge" (now `dfn-rdf-graph-merge`) and duplicate definitions for "RDF dataset" directly to the branch.

pfps:
This looks to be ready to go?  Any objections?

afs:
Do we want the title to be "Definition of an RDF dataset" when the definition is in RDF concepts? 

It could called something like "Description of an [RDF Dataset](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-dataset)"; alternatively, consider it a parallel definition and keep as it currently "Definition of an RDF dataset".

pfps:
Perhaps "Formal Definition" would work?

afs:
> Perhaps "Formal Definition" would work?

As in a "squiggly mathy definition"? :smile: . 

The question will then be why is the merge definition not "formal"?

The current "Definition of an RDF dataset" is better than "Formal Defintion" IMO.

Puts put it in as-is pending other suggestions.

pfps:
I think the right thing is to completely defer to RDF Concepts for the definition of an RDF dataset.  This means adding it to the imported terminology section at the beginning and then making the "definition" in the new section just an alternative way of saying that definition.

pfps:
The mess with the HTML PR needs to be resolved before this PR is updated, I think.

afs:
> The mess with the HTML PR needs to be resolved before this PR is updated, I think.

The HTML on `main` has been adjusted.

pfps:
So is this ready to go?

pfps:
With no negative comments and lots of time elapsed, I'm going to merge this shortly unless there are complaints.

afs:
Please merge.

(This is needed to address https://github.com/w3c/sparql-update/issues/51).

PR #52, opened 2024-11-01
Better HTML

domel:
This PR addresses several HTML structure and semantic issues to improve accessibility, maintainability, and adherence to best practices. The following corrections were implemented:

- Tables used in a non-semantic way: Replaced tables that were previously used solely for layout purposes with CSS styling. This change ensures that tables are used strictly for tabular data, enhancing semantic integrity and improving screen reader compatibility.
- Missing essential HTML elements: Added essential structural elements.
- Redundant HTML tags: Removed any duplicate or unnecessary HTML tags, streamlining the HTML structure. This cleanup reduces code complexity and improves performance by minimizing DOM size.
- Small fixes.

See https://github.com/w3c/rdf-semantics/issues/51

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/52.html" title="Last updated on Nov 4, 2024, 7:59 PM UTC (84888af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/52/e29d0aa...84888af.html" title="Last updated on Nov 4, 2024, 7:59 PM UTC (84888af)">Diff</a>

PR #55, opened 2024-12-02
New triple term semantics in the rdf-semantics spec

franconi:
Please check the _minimal_ extension I made to the `rdf-semantics` doc, with the new definitions as in the [RDF-star "working alternative baseline"](https://github.com/w3c/rdf-star-wg/wiki/RDF-star-%22working-alternative-baseline%22) spec we voted.
Note that I had to make a minor fix to the [RDF alternative semantics](https://github.com/w3c/rdf-star-wg/wiki/RDF-star-%22working-alternative-baseline%22) spec, in order to exclude that `rdf:type` becomes a `rdf:ReificationProperty`.
The `rdf-semantics` doc has been monotonically extended with the necessary new definitions about triple terms.
This version implements a minimal change, since it could be possible also to enhance the quality of the current text.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Dec 3, 2024, 4:46 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Frdf-semantics%2F06b2096d26cf76cab6835bad430535df3210a6e2%2Fspec%2Findex.html%3FisPreview%3Dtrue)

```
Timed out after waiting 30000ms
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/rdf-semantics%2355.)._
</details>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1704,6 +1704,7 @@
             <td class="othertable" style="background-color: Cyan;color: Black">sss aaa ttt <code>.</code><br/> for ttt a triple term</td>
             <td class="othertable" style="background-color: Cyan;color: Black">ttt <code>rdf:type</code> <code>rdf:TripleTerm .</code></td>
           </tr>
+          <tr>
             <td class="othertable" style="background-color: Cyan;color: Black"><dfn>G-rdfTT2</dfn></td>
             <td class="othertable" style="background-color: Cyan;color: Black">ttt aaa ooo <code>.</code><br/> for ttt a triple term</td>
             <td class="othertable" style="background-color: Cyan;color: Black">ttt <code>rdf:type</code> <code>rdf:TripleTerm .</code></td>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1700,7 +1700,6 @@
             <td class="othertable" style="background-color: Cyan;color: Black">aaa <code>rdf:type</code> <code>rdf:InverseReificationProperty .</code></td>
           </tr>
           <tr>
-          <tr>
             <td class="othertable" style="background-color: Cyan;color: Black"><dfn>G-rdfTT1</dfn></td>
             <td class="othertable" style="background-color: Cyan;color: Black">sss aaa ttt <code>.</code><br/> for ttt a triple term</td>
             <td class="othertable" style="background-color: Cyan;color: Black">ttt <code>rdf:type</code> <code>rdf:TripleTerm .</code></td>
--- a/spec/index.html
+++ b/spec/index.html
@@ -465,6 +465,7 @@
             <li>A mapping <em>IEXT</em> from <em>IP</em> into the powerset of <em>IR</em> x <em>IR</em>, i.e., the set of sets of pairs <em>&lt;x, y&gt;</em> with <em>x</em> and <em>y</em> in <em>IR</em>.</li>
             <li>A mapping <em>IS</em> from <em>IRIs</em> into (<em>IR</em> union <em>IP</em>).</li>
             <li>A partial mapping <em>IL</em> from literals into <em>IR</em>.</li>
+            <li><span style="background-color: Cyan;color: Black">6. An injective mapping RE from IR x IP x IR into IR, called the denotation of triple terms</span> </li>
           </ol>
         </div>
       </section>
@@ -506,15 +507,18 @@
         from its set-theoretic extension.
         A similar technique is used in the ISO/IEC Common Logic standard [[?ISO24707]].
       </p>
-      <p>The referent of a ground RDF graph in a simple interpretation <em>I</em> is then given by the following rules,
+      <p><span style="background-color: Cyan;color: Black"><s>The referent of a ground RDF graph in a simple interpretation <em>I</em> is then given by the following rules,
         where the interpretation is also treated as a function from expressions (<a>names</a>, <a>triples</a> and <a>graphs</a>) to elements of the universe and truth values:
-      </p>
+      </s></span></p>
+      <p><span style="background-color: Cyan;color: Black">The referent of a ground RDF graph in a simple interpretation I is given by an <i>interpretation</i> function I from expressions (<a>names</a>, <a>triples</a> and <a>graphs</a>) to elements of the universe and truth values, defined as follows:</span></p>
       <section class="notoc">
         <h6 class="defcap">Semantic Conditions for Ground Graphs</h6>
         <div class="semantictable">
           <p>If <em>E</em> is a literal, then <em>I(E) = IL(E)</em></p>
           <span class="space"></span>
           <p>If <em>E</em> is an IRI, then <em>I(E) = IS(E)</em></p>
+          <span class="space"></span>
+          <p><span style="background-color: Cyan;color: Black">if <em>E</em> is a Triple Term then <em>I(E) = RE(I(E.s), I(E.p), I(E.o))</em><br> where we write the subject, predicate, object of <em>E</em> as <em>E.s, E.p, E.o</em>, respectively.</span></p>
           <span class="space"></span>
           <p>If <em>E</em> is a ground triple `s p o.`, <em>I(p)</em>
             is in <em>IP</em> and the pair &lt;<em>I(s)</em>, <em>I(o)</em>&gt;
@@ -957,6 +961,9 @@
           <li><code>rdf:subject</code></li>
           <li><code>rdf:predicate</code></li>
           <li><code>rdf:object</code></li>
+          <li><code><span style="background-color: Cyan;color: Black">rdf:TripleTerm</span></code></li>
+          <li><code><span style="background-color: Cyan;color: Black">rdf:ReificationProperty</span></code></li>
+          <li><code><span style="background-color: Cyan;color: Black">rdf:InverseReificationProperty</span></code></li>
           <li><code>rdf:first</code></li>
           <li><code>rdf:rest</code></li>
           <li><code>rdf:value</code></li>
@@ -985,6 +992,36 @@
             <em>IEXT(I(<code>rdf:type</code>))</em> if and only if <em>x</em> is in the 
             <a>value space</a> of <em>I(aaa)</em>.
           </p>
+          <span class="space"></span>
+          
+          <p id="rdfsemcond1"><span style="background-color: Cyan;color: Black">
+            if <em>[I+A](E) = TRUE</em> and <em>E.s</em> is a triple term, then
+            <em><[I+A](E.s), I(<code>rdf:TripleTerm</code>)></em> is in <em>IEXT(I(<code>rdf:type</code>))</em>.
+          </span></p>
+          <span class="space"></span>
+      
+          <p id="rdfsemcond1"><span style="background-color: Cyan;color: Black">
+            if <em>[I+A](E) = TRUE</em> and <em>E.o</em> is a triple term, then
+            <em><[I+A](E.o), I(<code>rdf:TripleTerm</code>)></em> is in <em>IEXT(I(<code>rdf:type</code>))</em>.
+          </span></p>
+          <span class="space"></span>
+       
+          <p id="rdfsemcond1"><span style="background-color: Cyan;color: Black">
+            if <em>[I+A](E) = TRUE</em> and 
+            <em><[I+A](E.o), [I+A](rdf:TripleTerm)></em> is in <em>IEXT(I(<code>rdf:type</code>))</em> and
+            <em>E.p ≠ <code>rdf:type</code></em>, then
+            <em><[I+A](E.p), I(<code>rdf:ReificationProperty</code>)></em> is in <em>IEXT(I(<code>rdf:type</code>))</em>.
+          </span></p>
+          <span class="space"></span>
+       
+          <p id="rdfsemcond1"><span style="background-color: Cyan;color: Black">
+            if <em>[I+A](E) = TRUE</em> and
+            <em><[I+A](E.s), [I+A](rdf:TripleTerm)></em> is in <em>IEXT(I(<code>rdf:type</code>))</em>
+            and <em>E.p ≠ <code>rdf:type</code></em>, then
+            <em><[I+A](E.p), I(<code>rdf:InverseReificationProperty</code>)></em> is in <em>IEXT(I(<code>rdf:type</code>))</em>.
+          </span></p>
+
+          
         </div>
       </section>
       <p>and <a>satisfies</a> every triple in the following infinite set:</p>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1619,6 +1680,38 @@ <h2>Entailment rules (Informative)</h2>
...
+            <td class="othertable" style="background-color: Cyan;color: Black">sss aaa ttt <code>.</code><br/> ttt <code>rdf:type</code> <code>rdf:TripleTerm .</code><br/>for aaa ≠ <code>rdf:type</code></td>
+            <td class="othertable" style="background-color: Cyan;color: Black">aaa <code>rdf:type</code> <code>rdf:ReificationProperty .</code></td>
+          </tr>
+          <tr>
+            <td class="othertable" style="background-color: Cyan;color: Black"><dfn>G-rdfREIF2</dfn></td>
+            <td class="othertable" style="background-color: Cyan;color: Black">ttt aaa ooo <code>.</code><br/> ttt <code>rdf:type</code> <code>rdf:TripleTerm .</code><br/>for aaa ≠ <code>rdf:type</code></td>
+            <td class="othertable" style="background-color: Cyan;color: Black">aaa <code>rdf:type</code> <code>rdf:InverseReificationProperty .</code></td>
+          </tr>
+          <tr>
+          <tr>
```
gkellogg, line 1703:
```suggestion
          <tr>
```

Doubled table row start.

spec/index.html
```diff
@@ -1619,6 +1680,38 @@ <h2>Entailment rules (Informative)</h2>
...
+            <td class="othertable" style="background-color: Cyan;color: Black"><dfn>G-rdfTT1</dfn></td>
+            <td class="othertable" style="background-color: Cyan;color: Black">sss aaa ttt <code>.</code><br/> for ttt a triple term</td>
+            <td class="othertable" style="background-color: Cyan;color: Black">ttt <code>rdf:type</code> <code>rdf:TripleTerm .</code></td>
+          </tr>
```
gkellogg, line 1707:
```suggestion
          </tr>
          <tr>
```

This is where the opening row should be.

gkellogg (changes requested):
There are a bunch of markup errors, which is what keeps this from building properly.

spec/index.html
franconi: fixed

gkellogg (commented):
Generally, I think the PR Preview "diff" accessor (which seems broken right now) is the best way for people to see the changes for this PR rather than using permanent markup.

ReSpec also has some support for the `<ins>` and `<del>` elements, but that's typically used for flagging post-REC changes for consideration for subsequent normative changes. I am wary of using such markup, though.

gkellogg (approved):

spec/index.html
```diff
@@ -465,6 +465,7 @@ <h6 class="defcap">Definition of a Simple Interpretation</h6>
             <li>A mapping <em>IEXT</em> from <em>IP</em> into the powerset of <em>IR</em> x <em>IR</em>, i.e., the set of sets of pairs <em>&lt;x, y&gt;</em> with <em>x</em> and <em>y</em> in <em>IR</em>.</li>
             <li>A mapping <em>IS</em> from <em>IRIs</em> into (<em>IR</em> union <em>IP</em>).</li>
             <li>A partial mapping <em>IL</em> from literals into <em>IR</em>.</li>
+            <li><span style="background-color: Cyan;color: Black">6. An injective mapping RE from IR x IP x IR into IR, called the denotation of triple terms</span> </li>
```
afs, line 468:
```suggestion
            <li><span style="background-color: Cyan;color: Black">An injective mapping RE from IR x IP x IR into IR, called the denotation of triple terms</span> </li>
```
Remove number - that is generated by the `<li>`.

spec/index.html
```diff
@@ -1002,6 +1039,7 @@ <h6 class="defcap">RDF Axioms</h6>
             rdf:nil       rdf:type rdf:List .
             rdf:_1        rdf:type rdf:Property .
             rdf:_2        rdf:type rdf:Property .
+            <span style="background-color: Cyan;color: Black">rdf:reifies   rdf:type rdf:ReificationProperty .</span>
             ...
```
afs, line 1043:
```suggestion
            <span style="background-color: Cyan;color: Black">rdf:reifies   rdf:type rdf:ReificationProperty .</span>
            rdf:_1        rdf:type rdf:Property .
            rdf:_2        rdf:type rdf:Property .
            ...
```
Move new line up so the `...` remain directly after `rdf:_2`.

spec/index.html
```diff
@@ -1110,6 +1148,29 @@ <h4>Patterns of RDF entailment (Informative)</h4>
...
+             <tr>
+               <td class="othertable" style="background-color: Cyan;color: Black"><dfn>rdfREIF1</dfn></td>
+               <td class="othertable" style="background-color: Cyan;color: Black">sss aaa ttt <code>.</code><br/> for ttt a triple term</td>
+               <td class="othertable" style="background-color: Cyan;color: Black">aaa <code>rdf:type</code> <code>rdf:ReificationProperty .</code><br/>sss aaa _:nnn <code>.</code><br/>_:nnn <code>rdf:type</code><code> rdf:TripleTerm </code><code>.</code></td>
+             </tr>
+           </tbody>
+         </table>
+          
+          
+
```
afs, line 1173:
```suggestion
          
```

afs (commented):
Should there be changes in section 9 / RDFS Axiomatic Triples?

TallTed:
It seems this (#55) should wait until the impact of #52 is fully addressed, whether by reversion or otherwise. /cc @gkellogg @pfps @domel

spec/index.html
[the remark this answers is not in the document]
franconi: The "Cyan" background coloring is only meant at this stage to emphasise clearly the few additions I've have made - all focussed on the semantics and not on the wording. Once everybody agrees, I will remove the styling.
Maybe the next time I will learn how to achieve the same goal in a cleaner way...

franconi:
> Should there be changes in section 9 / RDFS Axiomatic Triples?



> Should there be changes in section 9 / RDFS Axiomatic Triples?

I am not sure whether the WG agrees on the RDFS extension with rdfs: states. You tell me :-)

spec/index.html
```diff
@@ -506,16 +507,19 @@ <h6 class="defcap">Definition of a Simple Interpretation</h6>
         from its set-theoretic extension.
         A similar technique is used in the ISO/IEC Common Logic standard [[?ISO24707]].
       </p>
-      <p>The referent of a ground RDF graph in a simple interpretation <em>I</em> is then given by the following rules,
+      <p><span style="background-color: Cyan;color: Black"><s>The referent of a ground RDF graph in a simple interpretation <em>I</em> is then given by the following rules,
```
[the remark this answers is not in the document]
pchampin: +1 I appreciate @franconi 's efforts to help others review the changes, but mixing this metadata (what has change) with the data (the content of the spec) seems like a suboptiomal way to do this...

This is exactly what pr-preview is supposed to do (although it seems to be broken right now...).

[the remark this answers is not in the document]
pchampin: Also, I found the previous wording "where the interpretation is also treated as a function" less confusing than the new wording...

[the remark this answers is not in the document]
franconi: Few points:
1) could you teach me how to use/enable pr-preview?
2) Due to the parallel radical changes of the formatting in the main branch, it is hard to compare the proposed changes wrt the current main state. That's why I have created a self-standing version (see point 3 below); once a final agreement is found, I will make sure to create a compatible version for merging.
3) For everybody: in order to see the proposed changes, that is how I would do: Download all <https://github.com/franconi/rdf-semantics-altsem/tree/triple-terms-semantics/spec> and open index.html in a browser.

spec/index.html
```diff
@@ -957,6 +961,9 @@ <h6 class="defcap"><dfn>RDF vocabulary</dfn></h6>
           <li><code>rdf:subject</code></li>
           <li><code>rdf:predicate</code></li>
           <li><code>rdf:object</code></li>
+          <li><code><span style="background-color: Cyan;color: Black">rdf:TripleTerm</span></code></li>
```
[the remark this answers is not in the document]
franconi: The domain of discourse of such a class is exactly _triple terms_ (as **syntactic** objects). E.g., in an unstar procedure, the generated "middle" bnode would be of type _triple term_, since it is exactly what it is meant to denote.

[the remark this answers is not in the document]
franconi: It is possible that in some future WG, while defining the semantics for named graphs, the type Triple should be used for actual triples in the graph, not just triple terms.

spec/index.html
```diff
@@ -985,6 +992,36 @@ <h6 class="defcap">RDF Semantic Conditions</h6>
...
+            <em>&lt;[I+A](E.o), [I+A](rdf:TripleTerm)&gt;</em> is in <em>IEXT(I(<code>rdf:type</code>))</em> and
+            <em>E.p ≠ <code>rdf:type</code></em>, then
+            <em>&lt;[I+A](E.p), I(<code>rdf:ReificationProperty</code>)&gt;</em> is in <em>IEXT(I(<code>rdf:type</code>))</em>.
+          </span></p>
+          <span class="space"></span>
+       
+          <p id="rdfsemcond7"><span style="background-color: Cyan;color: Black">
+            if <em>[I+A](E) = TRUE</em> and
+            <em>&lt;[I+A](E.s), [I+A](rdf:TripleTerm)&gt;</em> is in <em>IEXT(I(<code>rdf:type</code>))</em>
+            and <em>E.p ≠ <code>rdf:type</code></em>, then
```
pchampin, line 1020: I understand that, without this exception, `rdf:type` would automatically become an InverseReificationProperty (because of the rules above).

Yet, would that be so bad?

spec/index.html
```diff
@@ -1110,6 +1148,27 @@ <h4>Patterns of RDF entailment (Informative)</h4>
...
+           <tbody style="background-color: Cyan;color: Black">
+             <tr style="background-color: Cyan;color: Black">
+               <th > </th>
+               <th ><strong>if <em>S</em> contains</strong></th>
+               <th ><strong>then <em>S</em> RDF entails</strong></th>
+             </tr>
+             <tr>
+               <td class="othertable" style="background-color: Cyan;color: Black"><dfn>rdfREIF1</dfn></td>
+               <td class="othertable" style="background-color: Cyan;color: Black">sss aaa ttt <code>.</code><br/> for ttt a triple term</td>
+               <td class="othertable" style="background-color: Cyan;color: Black">aaa <code>rdf:type</code> <code>rdf:ReificationProperty .</code><br/>sss aaa _:nnn <code>.</code><br/>_:nnn <code>rdf:type</code><code> rdf:TripleTerm </code><code>.</code></td>
```
pchampin, line 1167: In my opinion, this should be restricted to `rdf:reifies` only.
Other ReificationProperties may or may not have range `rdf:TripleTerm`.

ReificationProperty was proposed as a mere "flag" to allow people to detect errors (IF they consider this as an error). With this inference rule, it goes beyond flagging, and has other important consequences.

franconi: I tendo to agree here.

spec/index.html
franconi: Yes, it would be bad for several reasons, not least, as you say, because "ReificationProperty was proposed as a mere "flag" to allow people to detect errors", while here there is no error.

spec/index.html
```diff
@@ -985,6 +992,36 @@ <h6 class="defcap">RDF Semantic Conditions</h6>
...
+          <p id="rdfsemcond5"><span style="background-color: Cyan;color: Black">
+            if <em>[I+A](E) = TRUE</em> and <em>E.o</em> is a triple term, then
+            <em>&lt;[I+A](E.o), I(<code>rdf:TripleTerm</code>)&gt;</em> is in <em>IEXT(I(<code>rdf:type</code>))</em>.
+          </span></p>
+          <span class="space"></span>
+       
+          <p id="rdfsemcond6"><span style="background-color: Cyan;color: Black">
+            if <em>[I+A](E) = TRUE</em> and 
+            <em>&lt;[I+A](E.o), [I+A](rdf:TripleTerm)&gt;</em> is in <em>IEXT(I(<code>rdf:type</code>))</em> and
+            <em>E.p ≠ <code>rdf:type</code></em>, then
```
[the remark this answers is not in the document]
franconi: see below.

spec/index.html
franconi: if the user writes something like x rdf:type <<( s p o )>>, then flagging rdf:type as a reification property IS NOT appropriate, since rdf:type is meant to have a specific meaning (made explicit in RDFS), and the user is expected to respect that intended meaning, and it is not our job to emphasizes that they are making a mistake.

doerthe:
I am fine with the request in general (modulo some details I will report separately), but wonder why `rdf:type` can't be a `rdf:ReificationProperty`. How do others see that? How do we want to use that class in general?

And of course, I dislike the injectivity and am afraid that this complicates things without bringing benefit ...
(to add my own "Ceterum censeo Carthaginem esse delendam" ;) )

spec/index.html
doerthe: I see that my comment below is already made by PA. I also do not see the problem. So, we would want that reification properties have ONLY triple terms in object position or what is our goal here?

spec/index.html
pchampin: actually, I misread this inference rule. It does *not* imply that all ReificationProperties would all have range `rdf:TripleTerm`, and I have no issue with it.

spec/index.html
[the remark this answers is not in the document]
gkellogg: > 1. could you teach me how to use/enable pr-preview?

PR Preview is a service that is enabled in W3C repositories with a `.pr-preview.json` configuration file. It automatically builds the preview, and diff and puts links to them at the end of the first comment on the PR. It updates with every new commit. Sometimes the tooling breaks, though. In this case githack.com can be useful.

pfps:
A problem with this PR is that it includes a lot of formatting changes, which make it very hard to determine what really has changed.

PR #100, opened 2025-03-06
remove informative appendix on reification, etc.

### LINKED ISSUE #85: remove informative section about reification, collections, and containers from Semantics
https://www.w3.org/TR/rdf12-semantics/#non_semantics  is non-normative.  Its only semantics content is to point out some non-entailments.   

But anyone who is confused as to how these IRIs don't work isn't going to get anything from the semantics bits of this section.

So I suggest that the section be removed.  Any useful non-semantics content could be moved elsewhere.

pfps:
Fixes #85

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/100.html" title="Last updated on Mar 6, 2025, 1:12 PM UTC (690944d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/100/d6f7b9d...690944d.html" title="Last updated on Mar 6, 2025, 1:12 PM UTC (690944d)">Diff</a>


doerthe:
I agree with the change but we should combine it with an issue somewhere else to make sure that the informal appendix is explained somewhere. Or is that already the case?

niklasl:
I'm not sure we agreed to remove _all_ of the appendices, but i don't mind _moving_ them; possibly to different places.

On [the call of 2025-02-28](https://www.w3.org/2025/02/28-rdf-star-minutes.html) we talked about whether/what to move. Some notes:

* The Primer shouldn't mention legacy things or go much into advanced practice, so it's not very suitable.
* If moved, we don't necessarily need to have links from Semantics, since these are "forever" in 1.1 Semantics.
* The section in Semantics sort of served to mention that while these terms exist in the RDF vocabulary, they *don't* have a formal semantics...

I tentatively suggest:
* Collections (i.e. lists) have syntax in all W3C RDF formats, and are _extensively_ used, including in OWL. That _could_ go into concepts. An alternative is to add it to RDF Schema, and reference it from concepts.
* Containers are reasonably legacy/archaic (some users might disagree). Perhaps it can be moved to an appendix of RDF Schema, with some "soft" deprecation language (in favour of collections)?
*  The classic reification section could go into [the Note](https://github.com/w3c/rdf-star-wg/issues/150) I have an action to draft.

TallTed:
```
    <p>There is no way in RDF to assert
      that a container contains only a fixed number of members. This is a
      reflection of the fact that it is always consistent to add a triple
      to a graph asserting a membership property of any container. And
      finally, there is no built-in assumption that an RDF container has
      only finitely many members.</p>
```

This is an important difference between containers and collections. To my mind it is sufficient reason to have both, and not treat either as "legacy/archaic". They are different constructs, and both have reasons to exist, or they would not both have been specified in the first place.

```
    <p>Also, RDF imposes no 'well-formedness' conditions on the use of this
      vocabulary, so that it is possible to write RDF graphs which assert
      the existence of highly peculiar objects such as lists with forked
      or non-list tails, or multiple heads:</p>
[...]
    <p>It is also possible to write a set of triples which under-specify a collection
      by failing to specify its <code>rdf:rest</code> property value.</p>
```

I do think there might be a way to specify a hybrid construct which supports both the infinitely expandable nature of a container *and* the forkable collection (which forking does mean that there may not be an `rdf:nil` on every branch, potentially making an infinitely expandable collection). Such a hybrid contruct *could* replace both these existing constructs, but the existing constructs MUST be supported for some indefinite (perhaps perpetual) period even if they're being deprecated, because we have no way of knowing how much data exists with these "legacy/archaic"/deprecated constructs.

niklasl:
I was surprised to see this merged without any approvals. IIRC, there where some thoughts from @pchampin regarding if _all_ could be removed/moved? In any case, we need follow-up issues to put (some of) this somewhere else. See my [suggested destinations](https://github.com/w3c/rdf-semantics/pull/100#issuecomment-2706915323) above.

PR #126, opened 2025-03-27
Some improved language, punctuation, and markup in rdf-semantics

TallTed:
***
<a href="https://pr-preview.s3.amazonaws.com/TallTed/rdf-semantics/pull/126.html" title="Last updated on Jul 31, 2025, 2:55 PM UTC (d691856)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/126/3c56850...TallTed:d691856.html" title="Last updated on Jul 31, 2025, 2:55 PM UTC (d691856)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -285,26 +285,26 @@
     <p>Two graphs are <a data-cite="RDF12-CONCEPTS#dfn-graph-isomorphism">isomorphic</a> when each maps into the other by a 1:1 mapping on blank nodes. Isomorphic graphs are mutual instances with an invertible instance
       mapping. As blank nodes have no particular identity beyond their location in a graph, we will often treat isomorphic graphs as identical.</p>
 
-    <p >An RDF graph is <dfn>lean</dfn> if it has no instance which is
+    <p>An RDF graph is <dfn>lean</dfn> if it has no instance which is
       a <a>proper subgraph</a> of itself. Non-lean graphs have internal redundancy
-      and express the same content as their lean <a>subgraphs</a>. For example, the graph</p>
+      and express the same content as their lean <a>subgraphs</a>. For example, the graph&nbsp;&mdash;</p>
 
     <pre class="example" title="Non-lean graph">
       ex:a ex:p _:x .
       _:y ex:p _:x .
     </pre>
 
-    <p >is not lean, but</p>
+    <p>&mdash;&nbsp;is not lean, but&nbsp;&mdash;</p>
 
     <pre class="example" title="Lean graph">
       ex:a ex:p _:x .
       _:x ex:p _:x .
     </pre>
 
-    <p>is lean. A <a>ground RDF graph</a> is lean. </p>
+    <p>&mdash;&nbsp;is lean. A <a>ground RDF graph</a> is lean.</p>
     
   <section id="unions_merges">
-    <h3>Shared blank nodes, unions and merges</h3>
+    <h3>Shared blank nodes, unions, and merges</h3>
 
     <p>Graphs share <a>blank nodes</a> only if they are derived from graphs
       described by documents or other structures (such as an RDF dataset)
@@ -313,7 +313,8 @@
       graph are the same as the blank nodes coming from other downloads of
       the same document or from the same <a>RDF source</a>.</p>
 
-    <p> RDF applications which manipulate concrete syntaxes for RDF which use <a data-cite="RDF12-CONCEPTS#dfn-blank-node-identifier">blank node identifiers</a>
+    <p> RDF applications which manipulate concrete syntaxes for RDF which use
+      <a data-cite="RDF12-CONCEPTS#dfn-blank-node-identifier">blank node identifiers</a>
       should take care to keep track of the identity of the blank nodes they identify.
       Blank node identifiers often have a local scope,
       so when RDF from different sources is combined,
@@ -324,11 +325,12 @@
       to identify a blank node, but unless these documents are in a shared identifier scope
       or are derived from a common source,
       the occurrences of "<code>_:x</code>" in one document
-      will identify a different blank node than the one in the graph described by the other document.
+      will identify a different blank node than the occurrences of "<code>_:x</code>"
+      in the graph described by the other document.
       When graphs are formed by combining RDF from multiple sources,
-      it may be necessary to <dfn class="no-export lint-ignore">standardize</dfn> apart the blank node identifiers
-      by replacing them by others which do not occur in the other document(s).
-      For example, the two graphs represented  by the following texts:</p>
+      it may be necessary to <dfn class="no-export lint-ignore">standardize</dfn> the blank node identifiers
+      by replacing them with other identifiers which do not occur in the other document(s).
+      For example, the two graphs represented  by the following texts&nbsp;&mdash;</p>
 
     <p><code>ex:a ex:p _:x . </code><br/><br/>
       <img src="RDF12SemanticsDiagrams/example1.svg" alt="Graph 1" /></p>
@@ -336,27 +338,29 @@
     <p><code>ex:b ex:q _:x . </code><br/><br/>
       <img src="RDF12SemanticsDiagrams/example2.svg" alt="Graph 2" ></p>
 
-    <p>contain four nodes. Their union would therefore also contain four nodes:</p>
+    <p>&mdash;&nbsp;contain four nodes. Their union would therefore also contain four nodes&nbsp;&mdash;</p>
 
     <p><img src="RDF12SemanticsDiagrams/example4.svg" alt="Union Graph"></p>
 
-    <p>However, the document formed by simply concatenating these textual surface representations:</p>
+    <p>However, the document formed by simply concatenating these textual surface representations&nbsp;&mdash;</p>
 
     <p><code>ex:a ex:p _:x .<br/>
     ex:b ex:q _:x .</code><br/></p>
 
-    <p>describes a graph containing three nodes:</p>
+    <p>&mdash;&nbsp;describes a graph containing three nodes&nbsp;&mdash;</p>
 
     <p><img src="RDF12SemanticsDiagrams/example3.svg" alt="Incorrect Union Graph"></p>
 
-    <p> since the two occurrences of the blank node identifier "<code>_:x</code>" occurring in a common identifier scope identify the same blank node. The four-node union of these two graphs is more properly described by a surface form such as:</p>
+    <p>&mdash;&nbsp;since the two occurrences of the blank node identifier "<code>_:x</code>
+      occurring in a common identifier scope identify the same blank node. The four-node union
+      of these two graphs is more properly described by a surface form such as&nbsp;&mdash;</p>
 
     <p><code>ex:a ex:p _:x1 .<br/>
       ex:b ex:q _:x2 .</code></p>
 
-    <p>in which the blank node identifiers have been <a data-lt="standardize">standardized</a>
+    <p>&mdash;&nbsp;in which the blank node identifiers have been <a data-lt="standardize">standardized</a>
       apart to avoid conflating the distinct blank nodes.
-      (The particular blank node identifiers used have no significance, only that they are distinct.)</p>
+      (The particular blank node identifiers used have no significance; it matters only that they are distinct.)</p>
 
     <p>It is possible for two or more graphs to share a blank node,
       for example if they are <a>subgraphs</a> of a single larger graph or derived from a common source.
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -324,39 +325,42 @@ <h3>Shared blank nodes, unions and merges</h3>
...
       or are derived from a common source,
       the occurrences of "<code>_:x</code>" in one document
-      will identify a different blank node than the one in the graph described by the other document.
+      will identify a different blank node than the occurrences of "<code>_:x</code>"
+      in the graph described by the other document.
       When graphs are formed by combining RDF from multiple sources,
-      it may be necessary to <dfn class="no-export lint-ignore">standardize</dfn> apart the blank node identifiers
-      by replacing them by others which do not occur in the other document(s).
-      For example, the two graphs represented  by the following texts:</p>
+      it may be necessary to <dfn class="no-export lint-ignore">standardize</dfn> the blank node identifiers
```
doerthe, line 331: "standardization apart" is a fixed term (Logic Programming) and I would therefore not change that

spec/index.html
```diff
@@ -324,39 +325,42 @@ <h3>Shared blank nodes, unions and merges</h3>
...
+      by replacing them with other identifiers which do not occur in the other document(s).
+      For example, the two graphs represented  by the following texts&nbsp;&mdash;</p>
```
TallTed, line 333:
```suggestion
      For example, the two graphs represented by the following texts&nbsp;&mdash;</p>
```

spec/index.html
afs: RDF Semantics is not defining "standardize apart".

Just remove the <dfn> to leave the text: 

"it may be necessary to standardize apart the blank node identifiers"

TallTed: I think @afs advises well (hence part of this text suggestion), but based on what I found in documentation [Google showed me when I asked "what does "Standardize apart" mean?"](https://www.google.com/search?q=what+does+%22Standardize+apart%22+mean%3F), I've reordered some words, which I think makes the sentence more readable. (We might consider adding an external link to some Logic Programming text which goes further into this subject, which I believe we have not made a prerequisite for readers of our specifications.)
```suggestion
      it may be necessary to standardize the blank node identifiers apart,
```

afs: The operation name is "standardize apart" as @doerthe notes.
The google hits has it as the two words together each time.

For better or worse, the RDF Semantics introduction starts as the document goes on:
"This document defines a model-theoretic semantics ..."

(and why I haven't done a technical review of RDF Semantics.)

TallTed: Any objection to *this*, which makes plain that it's a complex verb? I have spent some fruitless time searching for a page that clearly and simply defines this "operation", but all pages I've found treat it as preexisting knowledge, which is clearly a problematic assumption, as it was not preexisting for me. I would much rather entirely remove this term of art, and say the same thing with other words.
```suggestion
      it may be necessary to "standardize apart" the blank node identifiers
```

afs: I think it is OK but (1) I am not an editor of this document (2) experts in the field may have a view.

pfps:
I don't see any need to add m-dashes around Examples 1 and 2.  In fact, my view is that they are a distraction.

Although I prefer Oxford commas there is no reason to change to them in one place unless this is the only place where there is a missing Oxford comma.

TallTed:
> I don't see any need to add m-dashes around Examples 1 and 2. In fact, my view is that they are a distraction.

They serve as a connector between the sentence fragments preceding and following the code blocks, and those code blocks. Their presence makes clear that there was no accidental omission of other punctuation (whether period, comma, or otherwise). It is far more distracting to try to figure out what punctuation might have been left out, than it is to read the flow with the em-dashes.

> Although I prefer Oxford commas there is no reason to change to them in one place unless this is the only place where there is a missing Oxford comma.

So if there are 2 missing terminal commas, in a document which otherwise has them in all eligible positions, there's no reason to add them in those 2 locations? What if there are 3 missing? Why should *one* missing comma be added, but *two or more* missing commas be left that way?

spec/index.html
franconi: > _ping_ oh ye experts in the field (@doerthe @franconi @niklasl (are there other such experts among us?))! Please chime in on [my latest suggested change](https://github.com/w3c/rdf-semantics/pull/126#discussion_r2021363646)?

I think it is ok, as it uses the appropriate term.

pfps (changes requested):
No em-dashes, for starters.

spec/index.html
doerthe: Why not `<dfn class="no-export lint-ignore">standardize</dfn>` ?. There is a link to that later in the document. If we remove the dfn context here, we also need to remove the link (line 361). Alternatively, the dfn could be for "standardize apart"

TallTed: > [@doerthe] Alternatively, the `dfn` could be for "standardize apart"

I could go for that. In fact, [I did on March 28](https://github.com/w3c/rdf-semantics/pull/126#discussion_r2019116689).

> [@afs] RDF Semantics is not defining "standardize apart".

The `dfn` tag does not necessarily indicate that this document is defining a term for use in other documents. It can and often does indicate the location of the working definition being used within this document, which linkage is useful for readers unfamiliar with such niche terminology as "standardize apart".

TallTed:
There's an open change request on the PR, and the PR itself has multiple change requests on the document. I don't think the PR should be closed.

spec/index.html
gkellogg: You can use `dfn` with `data-cite` which will style and reference the term, but not define it within the document. Clicking on it would take you to the `data-cite` reference, which must be of the form `SPECREF#anchor`. But, not strictly necessary if people reading this would understand the terminology without an explicit reference.

TallTed:
Do you have similar objections to *all* non-ASCII characters? I haven't noticed any such, but perhaps I've overlooked them.

TallTed:
If the WG decision is to remove all em-dashes, then the whole text of all documents must be reviewed for such, and various rephrasing and/or repunctuating will be required.

The em-dashes of immediate concern here precede and/or follow codeblocks. The text that precedes such codeblocks can be rewritten such that a colon may replace the em-dash — (following an em-dash, he continued) but colons cannot replace these em-dashes without some rephrasing. Colons only belong at a point in a sentence where a period might go — not midphrase, as an em-dash may be. Different rephrasing is necessary for sentence fragments that are now both preceded and followed by an em-dash. These must be entirely rewritten to be complete sentences, instead of the few, or even one, words they are now.

TallTed:
OK, replace all em-dash (`—`, which I type with alt-shift-hyphen on macOS) with three hyphens (`---`), and all en-dash (`–`, which I type with alt-hyphen on macOS) with two hyphens (`--`). Hyphens are already hyphens. These are the traditional substitutions, used on mechanical typewriters.

Also — why do we care what other people will type, when they're not editing our documents?

gkellogg:
I actually like em-dash, and find it easy to type on my Mac. Dbut, unlike regular convention, I prefer whitespace around the em-dash.

spec/index.html
```diff
@@ -285,26 +285,26 @@ <h2>Notation and Terminology</h2>
...
-    <p >is not lean, but</p>
+    <p>In contrast, the graph in Example 2 <em>is</em> lean:</p>
 
     <pre class="example" title="Lean graph">
       ex:a ex:p _:x .
       _:x ex:p _:x .
     </pre>
 
-    <p>is lean. A <a>ground RDF graph</a> is lean. </p>
+    <p>A <a>ground RDF graph</a> is also lean.</p>
```
pfps, line 304: The also is unneeded and only interferes with the meaning of the sentence.

pfps, line 304:
```suggestion
    <p>A <a>ground RDF graph</a> is lean.</p>
```

spec/index.html
```diff
@@ -324,39 +325,46 @@ <h3>Shared blank nodes, unions and merges</h3>
       to identify a blank node, but unless these documents are in a shared identifier scope
       or are derived from a common source,
       the occurrences of "<code>_:x</code>" in one document
-      will identify a different blank node than the one in the graph described by the other document.
+      will identify a different blank node than the occurrences of "<code>_:x</code>"
+      in the graph described by the other document.
```
pfps, line 329:
```suggestion
      in the other document.
```

spec/index.html
```diff
@@ -324,39 +325,46 @@ <h3>Shared blank nodes, unions and merges</h3>
...
       When graphs are formed by combining RDF from multiple sources,
       it may be necessary to <dfn class="no-export lint-ignore">standardize</dfn> apart the blank node identifiers
-      by replacing them by others which do not occur in the other document(s).
-      For example, the two graphs represented  by the following texts:</p>
+      by replacing them with other identifiers which do not occur in the other document(s).
+      For example, the two graphs represented by the two texts below contain two nodes each,
```
pfps, line 333:
```suggestion
      For example, the two graphs for the two texts below contain two nodes each, 
```

pfps: It is better to use a vague relationship than to have different, not-supported-by-semantics relationships.

spec/index.html
```diff
@@ -324,39 +325,46 @@ <h3>Shared blank nodes, unions and merges</h3>
...
 
-    <p>However, the document formed by simply concatenating these textual surface representations:</p>
+    <p>However, if we simply concatenate these textual surface representations 
+      to form a new document, as shown below:</p>
 
     <p><code>ex:a ex:p _:x .<br/>
     ex:b ex:q _:x .</code><br/></p>
 
-    <p>describes a graph containing three nodes:</p>
+    <p>This new document describes a graph containing three nodes, because
```
pfps, line 352:
```suggestion
    <p>The grap for this new document contains three nodes, because
```

spec/index.html
```diff
@@ -324,39 +325,46 @@ <h3>Shared blank nodes, unions and merges</h3>
...
+      the two occurrences of the blank node identifier "<code>_:x</code>
+      now occur in a common identifier scope, and thus identify the same blank node,
+      as shown below:</p>
 
     <p><img src="RDF12SemanticsDiagrams/example3.svg" alt="Incorrect Union Graph"></p>
 
-    <p> since the two occurrences of the blank node identifier "<code>_:x</code>" occurring in a common identifier scope identify the same blank node. The four-node union of these two graphs is more properly described by a surface form such as:</p>
+    <p>The four-node union
+      of these two graphs is more properly described by a surface form such as the following:</p>
```
pfps, line 360:
```suggestion
      of these two graphs is instead:</p>
```

spec/index.html
```diff
@@ -324,39 +325,46 @@ <h3>Shared blank nodes, unions and merges</h3>
...
 
     <p><code>ex:a ex:p _:x1 .<br/>
       ex:b ex:q _:x2 .</code></p>
 
-    <p>in which the blank node identifiers have been <a data-lt="standardize">standardized</a>
-      apart to avoid conflating the distinct blank nodes.
-      (The particular blank node identifiers used have no significance, only that they are distinct.)</p>
+    <p>Here the blank node identifiers have been <a data-lt="standardize">standardized</a> apart
```
pfps, line 365:
```suggestion
    <p>Here the blank node identifiers have been <a data-lt="standardize">standardized apart</a>
```

spec/index.html
afs: This whole paragraph does rather dangle. It would be better somewhere else, maybe after the sentence:

"An RDF graph is lean if it has no instance which is a [proper subgraph](https://www.w3.org/TR/rdf12-semantics/#dfn-proper-subgraph) of itself."

spec/index.html
TallTed: Fine

spec/index.html
TallTed: You can only have the colon if it is preceded by a complete sentence, like the following:
```suggestion
      of these two graphs is instead that shown below:</p>
```

spec/index.html
TallTed: I would laugh if I hadn't been trying to get something like this accepted for weeks!

spec/index.html
TallTed: Is redundant, given #146.

pfps: This PR should not include the "also" even if another PR may fix it up later.

pfps (changes requested):
With the "also" removed this looks reasonable to me.

spec/index.html
TallTed, line 304:
```suggestion
    <p>A <a>ground RDF graph</a> is lean.</p>
```

spec/index.html
```diff
@@ -285,26 +285,26 @@ <h2>Notation and Terminology</h2>
     <p>Two graphs are <a data-cite="RDF12-CONCEPTS#dfn-graph-isomorphism">isomorphic</a> when each maps into the other by a 1:1 mapping on blank nodes. Isomorphic graphs are mutual instances with an invertible instance
       mapping. As blank nodes have no particular identity beyond their location in a graph, we will often treat isomorphic graphs as identical.</p>
 
-    <p >An RDF graph is <dfn>lean</dfn> if it has no instance which is
+    <p>An RDF graph is <dfn>lean</dfn> if it has no instance which is
       a <a>proper subgraph</a> of itself. Non-lean graphs have internal redundancy
```
TallTed, line 289:
```suggestion
      a <a>proper subgraph</a> of itself. A <a>ground RDF graph</a> is lean. Non-lean graphs have internal redundancy
```

spec/index.html
```diff
@@ -285,26 +285,26 @@ <h2>Notation and Terminology</h2>
...
-    <p >is not lean, but</p>
+    <p>In contrast, the graph in Example 2 <em>is</em> lean:</p>
 
     <pre class="example" title="Lean graph">
       ex:a ex:p _:x .
       _:x ex:p _:x .
     </pre>
 
-    <p>is lean. A <a>ground RDF graph</a> is lean. </p>
+    <p>A <a>ground RDF graph</a> is lean.</p>
```
TallTed, line 304:
```suggestion
```

pfps (approved):

PR #152, opened 2025-08-14
remove use of abstract syntax in text

pfps:
Change sole use of "abstract syntax" to components of RDF graphs.  This not only removes a problematic phrase but better describes what is going on.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/152.html" title="Last updated on Aug 23, 2025, 1:07 PM UTC (5d053f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/152/cfab1f2...5d053f3.html" title="Last updated on Aug 23, 2025, 1:07 PM UTC (5d053f3)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -242,7 +242,7 @@
       The statement "A = B" means that there is one entity which both expressions "A" and "B" denote. 
       Angle brackets &lt; x, y &gt; are used to indicate an ordered pair of x and y.</p>
 
-    <p>Throughout this document, <a>RDF graph</a>s and their components
+    <p>Throughout this document, <a>RDF graphs</a> and their components
       are written using the notational conventions of the Turtle syntax [[!RDF12-TURTLE]].
       The namespace prefixes <code>rdf:</code> <code>rdfs:</code> and <code>xsd:</code>
       are used as in the RDF Concepts specification [[!RDF12-CONCEPTS]],


spec/index.html
```diff
@@ -242,7 +242,7 @@ <h2>Notation and Terminology</h2>
       The statement "A = B" means that there is one entity which both expressions "A" and "B" denote. 
       Angle brackets &lt; x, y &gt; are used to indicate an ordered pair of x and y.</p>
 
-    <p>Throughout this document, <a>RDF graph</a>s and other fragments of RDF abstract syntax
+    <p>Throughout this document, <a>RDF graph</a>s and their components
```
TallTed, line 245:
```suggestion
    <p>Throughout this document, <a>RDF graphs</a> and their components
```

TallTed: Generally speaking, singular terms have plural forms (e.g., term, terms), and plural terms have singular forms (e.g., errata, erratum), which are typically treated as synonymous in ReSpec, with links back to the same definition/description. Part of the justification for this is that the markup and highlight read *much* more clearly and attractively when the full form is within the tag, whether the plural is regular and just adds an "s" (or "es" or similar) (and is usually handled automatically by ReSpec) or irregular and substantially changes the word or its end (and may require explicit aliasing in the ReSpec markup). So no, to my mind, it does not make more sense for the markup to be `<a>RDF graph</a>s and`; rather, it makes more sense for the markup to be `<a>RDF graphs</a> and`.

PR #144, opened 2025-07-17
No connection between propositions and facts in model-theoretic semantics

rat10:
Add clarification that annotations on propositions can't annotate facts according to the model-theoretic semantics, but that an operational semantics may nudge users towards assuming such a connection.

***
<a href="https://pr-preview.s3.amazonaws.com/rat10/rdf-semantics/pull/144.html" title="Last updated on Jul 24, 2025, 10:45 AM UTC (8505cea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/144/ca90378...rat10:8505cea.html" title="Last updated on Jul 24, 2025, 10:45 AM UTC (8505cea)">Diff</a>

PR #156, opened 2025-09-04
Better text in Section 5.3 with the purpose of relating triple terms and asserted triples.

franconi:
Following the various discussions, and specifically a discussion with @lisp, I propose a better text in Section 5.3 on propositions, facts, and asserted facts, with the purpose of relating triple terms and asserted triples.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -722,17 +722,23 @@
     
     <p>We define the <dfn>set of propositions</dfn> in an interpretation as follows:</p>
 
-	<p class="fact"> The set of propositions in an interpretation I is IPR(I) = {&nbsp;IT(x, y, z)&#65372;x is in IR, y is in IP, z is in IR&nbsp;}; we observe that a proposition is in the extension of <code>rdfs:Proposition</code>. </p>
+	<p class="fact"> The set of propositions in an interpretation I is IPR(I) = {&nbsp;IT(x, y, z)&#65372;x is in IR, y is in IP, z is in IR&nbsp;}. </p>
+	
+	<p>The denotation of a triple, being it a triple term or an asserted triple, is a proposition. Given the semantics of <a href="#rdfs_interpretations">RDFS Interpretations</a>, a proposition is in the extension of the class <code>rdfs:Proposition</code>.</p>
 
 	<p>We define the <dfn>set of facts</dfn> in an interpretation as follows:</p>
 
-	<p class="fact"> The set F of facts in an interpretation I is F(I) = {&nbsp;IT(x, y, z)&#65372;&lt;x, z&gt; is in IEXT(y)&nbsp;}. The set of facts is the set of propositions which are true in the interpretation. </p>
+	<p class="fact"> The set F of facts in an interpretation I is F(I) = {&nbsp;IT(x, y, z)&#65372;&lt;x, z&gt; is in IEXT(y)&nbsp;}. </p>
+	
+	<p>A fact in an interpretation is a proposition which is true in the interpretation. In other words, a fact in an interpretation is the proposition corresponding to the denotation of an asserted triple in the interpretation. </p>
 
 	<p>Given a blank node mapping, we define the <dfn>set of facts asserted by a graph</dfn> in an interpretation as follows:</p>
 
-	<p class="fact">Given a blank node mapping A, the set of all facts asserted by a graph G in an interpretation I is FEXT(G, I, A) = {&nbsp;IT(&nbsp;[I+A](s), I(p), [I+A](o)&nbsp;)&#65372;`s p o.` is in G&nbsp;}. We then observe that given a blank node mapping, the asserted facts of a graph with respect to an interpretation may not necessarily be among the facts of the interpretation.</p>
+	<p class="fact">Given a blank node mapping A, the set of all facts asserted by a graph G in an interpretation I is FEXT(G, I, A) = {&nbsp;IT(&nbsp;[I+A](s), I(p), [I+A](o)&nbsp;)&#65372;`s p o.` is in G&nbsp;}. </p>
 	
-	<p>We introduce a <dfn>general definition of satisfiability</dfn> of a graph in an interpretation as follows:</p>
+	<p>Given a blank node mapping and an interpretation, an asserted fact in a graph is the proposition corresponding to the denotation of a triple in the graph. These asserted facts may not necessarily be among the facts in the interpretation.</p>
+	
+	<p>We introduce a <dfn>general definition of satisfiability</dfn> of a graph in an interpretation, based on the intuition that a graph is satisfied by an interpretation if the asserted facts by the graph are among the facts of the interpretation.:</p>
 
 	<p class="fact">An interpretation (simply) satisfies a graph if and only if there exists a blank node mapping such that the facts asserted by the graph in the interpretation are among the facts of the interpretation.</p>
 	


franconi:
This should give an answer to [PR #144](https://github.com/w3c/rdf-semantics/pull/144), namely it clarifies the **connection between propositions and facts in model-theoretic semantics**.

franconi:
(please observe that I don't fully understand how to write in correct HTML the references to other parts of the doc)

spec/index.html
```diff
@@ -722,17 +724,23 @@ <h3>Properties of simple entailment and satisfiability</h3>
...
+	<p class="fact">Given a blank node mapping A, the set of all facts asserted by a graph G in an interpretation I is FEXT(G, I, A) = {&nbsp;IT(&nbsp;[I+A](s), I(p), [I+A](o)&nbsp;)&#65372;`s p o.` is in G&nbsp;}. </p>
 	
-	<p>We introduce a <dfn>general definition of satisfiability</dfn> of a graph in an interpretation as follows:</p>
+	<p>Given a blank node mapping and an interpretation, an asserted fact in a graph is the proposition corresponding to the denotation of a triple in the graph. These asserted facts may not necessarily be among the facts in the interpretation.</p>
+	
+	<p>We introduce a <dfn>general definition of satisfiability</dfn> of a graph in an interpretation, based on the intuition that a graph is satisfied by an interpretation if the asserted facts by the graph are among the facts of the interpretation.:</p>
 
 	<p class="fact">An interpretation (simply) satisfies a graph if and only if there exists a blank node mapping such that the facts asserted by the graph in the interpretation are among the facts of the interpretation.</p>
```
pchampin, line 745: Add a link to the definition of satisfies.
```suggestion
	<p class="fact">An interpretation (simply) <a>satisfies</a> a graph if and only if there exists a blank node mapping such that the facts asserted by the graph in the interpretation are among the facts of the interpretation.</p>
```

spec/index.html
```diff
@@ -1243,6 +1251,8 @@ <h2>RDFS Interpretations</h2>
     <em>Ext</em>ension in I) from IC to the set of subsets of IR.</p><p> A class may have an
     empty class extension. Two different classes can have the same class extension.
     The class extension of <code>rdfs:Class</code> contains the class <code>rdfs:Class</code>.</p>
+    
+    <p>RDFS introduces also the special class <code>rdfs:Proposition</code>, i.e. a resource whose extension is the set of propositions as defined in Section 5.3.</p>
```
pchampin, line 1255:
```suggestion
    <p>RDFS introduces also the special class <code>rdfs:Proposition</code>, i.e. a resource whose extension is the set of propositions as defined in <a href="#simple_entailment_properties">Section 5.3</a>.</p>
```

franconi:
> I suggested a few changes that I would strongly prefer to see applied, but generally this is a nice improvement.

Much better, thanks.

spec/index.html
```diff
@@ -1243,6 +1250,8 @@ <h2>RDFS Interpretations</h2>
     <em>Ext</em>ension in I) from IC to the set of subsets of IR.</p><p> A class may have an
     empty class extension. Two different classes can have the same class extension.
     The class extension of <code>rdfs:Class</code> contains the class <code>rdfs:Class</code>.</p>
+    
+    <p>RDFS introduces also the special class <code>rdfs:Proposition</code>, i.e. a resource whose extension is the set of propositions as defined in <a href="#simple_entailment_properties">Section 5.3</a>.</p>
```
niklasl, line 1254: Moving "also", linking the "set of propositions" and making a "smart" section link:
```suggestion
    <p>RDFS also introduces the special class <code>rdfs:Proposition</code>, i.e. a resource whose extension is the <a>set of propositions</a> as defined in [[[#simple_entailment_properties]]].</p>
```
(Note that the automatic section title is then expanded. The part beginning with "as defined in" could be in parentheses; or for that matter even omitted if the also linked "set of propositions" is enough on its own.)

franconi:
> I'm unclear what the term "corresponding" means, as used e.g. in the following sentence:
> 
> > Given a blank node mapping and an interpretation, an asserted fact in a graph is the proposition corresponding to the denotation of a triple in the graph. These asserted facts may not necessarily be among the facts in the interpretation. 

This sentence is just a rephrasing of the boxed definition.

> Also, I'm puzzled by the notion of an "asserted fact": so far I have assumed that asserting a triple (in a graph, one might add) claims it to be true (i.e. a fact), period. If a reader accepts that claim (and the graph containing it, one might add) was until now in my understanding outside the scope of RDF. Was I wrong all the time, or does this spec introduce a new way of discussing the truth of a statement by differentiating between ("universal"?) truth and truth in a graph? Sorry if this question itself sounds clumsy, but right now that's the best I can do. 

There are **facts**: these are the denotations of triples in general, independently on where they are used. Among them, there are the **asserted facts**, namely the denotation of triples which actually appear as triples in the graph (we call those asserted triples). Nothing more than that. Nothing about truth.

> Also, I'm still not properly understanding the ramifications of this construction on an assertion on a reifier: if there is a difference between a `fact` and an `asserted fact`, then from what I understand an assertion on a reifier might also constitute an assertion on a respective ("corresponding"?)`fact`, but not necessarily an `asserted fact`. Can you clarify that?

A fact is the denotation of any triple, asserted or not. So reification can apply to either kind of triple.

> Yet another issue: this construction seems to severely restrict the way in which properties alternatively to `rdf:reifies` can be defined. In particular it seems to make it impossible to define a property with weaker semantics than `rdf:reifes`, like e.g. `rdfx:citesButUnderNoCircumstancesEndorses` that creates a reference to a proposition, but never to its eventual assertion. So far this is just a question I have, but if that is indeed the case then I'm categorically opposed to this semantics. However, there would be an easy way out: keep the definition of propositions more general, and define a model-theoretic semantics of `rdf:reifies` to the desired effect.

No. See above.

> [Note: it took me a while to write this, and in the meantime some changes were made to the PR. I hope they have no impact on my questions here]

You are safe :-)

doerthe (commented):
I wanted to review before our meeting and therefore did not use the "suggest change" mode, sorry for that.

I like the changes, but  suggest to also use the names we defined (so when we talk about the facts if on interpretation, we also write F(I), etc...)

spec/index.html
```diff
@@ -504,6 +504,8 @@ <h2>Simple Interpretations</h2>
     the referent of the subject and object of any true triple will be in IR;
     so any IRI which occurs in a graph both as a predicate and as a subject or object
     will denote something in the intersection of IP and IR.</p>
+    
+  <p>We observe that the <code>rdf:reifies</code> property does not have any special semantic condition in RDF. Its intended meaning is thoroughly explained in Section 1.5 of the RDF Concepts document.</p>
```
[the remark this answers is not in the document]
pfps: I agree.  Even better would be to just remove this entirely.

spec/index.html
```diff
@@ -1243,6 +1250,8 @@ <h2>RDFS Interpretations</h2>
     <em>Ext</em>ension in I) from IC to the set of subsets of IR.</p><p> A class may have an
     empty class extension. Two different classes can have the same class extension.
     The class extension of <code>rdfs:Class</code> contains the class <code>rdfs:Class</code>.</p>
+    
+    <p>RDFS also introduces the special class <code>rdfs:Proposition</code>, i.e. a resource whose extension is the <a>set of propositions</a> as defined in [[[#simple_entailment_properties]]].</p>
```
pfps, line 1254: Is any other class marked as "special"?  If not, why single out this one?

franconi:
I've added a paragraph introducing rdf:reifies. Check it out.

niklasl:
> With the added stuff on rdf:reifies I take back my approval.
> 
> This is only acceptable to me if this is both necessary and sufficient to close all the issues related to rdf:reifies.

My hope is that it will, but agree that it would be useful to have confirmation on that (before merging this).

spec/index.html
```diff
@@ -504,6 +504,7 @@ <h2>Simple Interpretations</h2>
     the referent of the subject and object of any true triple will be in IR;
     so any IRI which occurs in a graph both as a predicate and as a subject or object
     will denote something in the intersection of IP and IR.</p>
+  <p>We observe that no IRI, not even those in the <code>rdf:</code> namespace, has any special semantic condition associated with it in a simple interpretation.</p>
```
TallTed, line 507: Shorter lines are better.
```suggestion
  <p>We observe that no IRI, not even those in the <code>rdf:</code> namespace,
  has any special semantic condition associated with it in a simple interpretation.</p>
```

spec/index.html
```diff
@@ -722,19 +723,25 @@ <h3>Properties of simple entailment and satisfiability</h3>
     
     <p>We define the <dfn>set of propositions</dfn> in an interpretation as follows:</p>
 
-	<p class="fact"> The set of propositions in an interpretation I is IPR(I) = {&nbsp;IT(x, y, z)&#65372;x is in IR, y is in IP, z is in IR&nbsp;}; we observe that a proposition is in the extension of <code>rdfs:Proposition</code>. </p>
+	<p class="fact"> The set of propositions in an interpretation I is IPR(I) = {&nbsp;IT(x, y, z)&#65372;x is in IR, y is in IP, z is in IR&nbsp;}. </p>
```
TallTed, line 726: It seems there should be some `code` or other formatting applied to some of these words or symbols... but I'm not sure which, so I'm not adding such. I am adding some `&nbsp;`.
```suggestion
	<p class="fact"> The set of propositions in an interpretation I is
	IPR(I)&nbsp;= {&nbsp;IT(x, y, z)&nbsp;&#65372;&nbsp;x&nbsp;is in IR,
	y&nbsp;is in IP, z&nbsp;is in IR&nbsp;}.</p>
```

spec/index.html
```diff
@@ -722,19 +723,25 @@ <h3>Properties of simple entailment and satisfiability</h3>
...
+	
+	<p>The denotation of a triple is a proposition, regardless of whether it is used as a triple term or an asserted triple. Given the semantics of <a href="#rdfs_interpretations">RDFS Interpretations</a>, a proposition is in the extension of the class <code>rdfs:Proposition</code>.</p>
```
TallTed, line 728: Applying @doerthe's [suggestion](https://github.com/w3c/rdf-semantics/pull/156/files#r2322556623)...
```suggestion
	<p>The denotation of a triple is a proposition, whether it is used as a triple
	term or an asserted triple. Under <a href="#rdfs_interpretations">RDFS
	Interpretations</a> (see below), a proposition is in the extension of the
	class <code>rdfs:Proposition</code>.</p>
```

spec/index.html
```diff
@@ -722,19 +723,25 @@ <h3>Properties of simple entailment and satisfiability</h3>
...
 
 	<p>We define the <dfn>set of facts</dfn> in an interpretation as follows:</p>
 
-	<p class="fact"> The set F of facts in an interpretation I is F(I) = {&nbsp;IT(x, y, z)&#65372;&lt;x, z&gt; is in IEXT(y)&nbsp;}. The set of facts is the set of propositions which are true in the interpretation. </p>
+	<p class="fact"> The set F of facts in an interpretation I is F(I) = {&nbsp;IT(x, y, z)&#65372;&lt;x, z&gt; is in IEXT(y)&nbsp;}. </p>
+	
+	<p>A fact in an interpretation is a proposition which is true in the interpretation. In other words, a fact in an interpretation is the proposition corresponding to the denotation of an asserted triple in the interpretation. </p>
```
TallTed, line 734:
```suggestion
	<p class="fact"> The set F of facts in an interpretation I is F(I)&nbsp;= {&nbsp;IT(x, y, z)&#65372;&lt;x, z&gt;&nbsp;is in IEXT(y)&nbsp;}. </p>
	
	<p>A fact in an interpretation is a proposition which is true in the
	interpretation. In other words, a fact in an interpretation is the proposition
	corresponding to the denotation of an asserted triple in the interpretation.</p>
```

spec/index.html
```diff
@@ -722,19 +723,25 @@ <h3>Properties of simple entailment and satisfiability</h3>
...
 
 	<p>Given a blank node mapping, we define the <dfn>set of facts asserted by a graph</dfn> in an interpretation as follows:</p>
```
TallTed, line 736:
```suggestion
	<p>Given a blank node mapping, we define the <dfn>set of facts asserted
	by a graph</dfn> in an interpretation as follows:</p>
```

spec/index.html
```diff
@@ -722,19 +723,25 @@ <h3>Properties of simple entailment and satisfiability</h3>
...
 
-	<p class="fact">Given a blank node mapping A, the set of all facts asserted by a graph G in an interpretation I is FEXT(G, I, A) = {&nbsp;IT(&nbsp;[I+A](s), I(p), [I+A](o)&nbsp;)&#65372;`s p o.` is in G&nbsp;}. We then observe that given a blank node mapping, the asserted facts of a graph with respect to an interpretation may not necessarily be among the facts of the interpretation.</p>
+	<p class="fact">Given a blank node mapping A, the set of all facts asserted by a graph G in an interpretation I is FEXT(G, I, A) = {&nbsp;IT(&nbsp;[I+A](s), I(p), [I+A](o)&nbsp;)&#65372;`s p o.` is in G&nbsp;}. </p>
```
TallTed, line 738:
```suggestion
	<p class="fact">Given a blank node mapping&nbsp;A, the set of all facts
	asserted by a graph&nbsp;G in an interpretation&nbsp;I is FEXT(G, I,
	A)&nbsp;= {&nbsp;IT(&nbsp;[I+A](s), I(p), [I+A](o)&nbsp;)&#65372;
	`s p o.`&nbsp;is in G&nbsp;}.</p>
```

spec/index.html
```diff
@@ -722,19 +723,25 @@ <h3>Properties of simple entailment and satisfiability</h3>
...
 	
-	<p>We introduce a <dfn>general definition of satisfiability</dfn> of a graph in an interpretation as follows:</p>
+	<p>Given a blank node mapping and an interpretation, an asserted fact in a graph is the proposition corresponding to the denotation of a triple in the graph. These asserted facts may not necessarily be among the facts in the interpretation.
+	Intuitively, this would only be the case if the interpretation satisfies the graph.
+	</p>
 
-	<p class="fact">An interpretation (simply) satisfies a graph if and only if there exists a blank node mapping such that the facts asserted by the graph in the interpretation are among the facts of the interpretation.</p>
+	<p class="fact">An interpretation (simply) <a>satisfies</a> a graph if and only if there exists a blank node mapping such that the facts asserted by the graph in the interpretation are among the facts of the interpretation.</p>
```
TallTed, line 744:
```suggestion
	<p class="fact">An interpretation&nbsp;I (simply) <a>satisfies</a> a graph&nbsp;G
	if and only if there exists a blank node mapping&nbsp;A such that the facts
	asserted by the graph in the interpretation&nbsp;FEXT(G,I,A) are a subset of
	the facts of the interpretation&nbsp;F(I).</p>
```

spec/index.html
```diff
@@ -1243,6 +1250,8 @@ <h2>RDFS Interpretations</h2>
     <em>Ext</em>ension in I) from IC to the set of subsets of IR.</p><p> A class may have an
     empty class extension. Two different classes can have the same class extension.
     The class extension of <code>rdfs:Class</code> contains the class <code>rdfs:Class</code>.</p>
+    
+    <p>RDFS also introduces the class <code>rdfs:Proposition</code>, which denotes a resource whose extension is exactly the <a>set of propositions</a> as defined in [[[#simple_entailment_properties]]]. The denotation of <code>rdfs:Proposition</code> also includes the range of the <code>rdf:reifies</code> property. A triple where the predicate is <code>rdf:reifies</code> and the object is therefore a proposition, is a <dfn>reifying triple</dfn>, and its subject is called a <dfn>reifier</dfn>.
```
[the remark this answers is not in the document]
TallTed: Just making @niklasl suggestion have shorter lines.
```suggestion
    <p>RDFS also introduces the class <code>rdfs:Proposition</code>, which denotes
    a resource whose extension is exactly the <a>set of propositions</a> as defined
    in [[[#simple_entailment_properties]]]. The denotation of
    <code>rdfs:Proposition</code> also includes the range of the
    <code>rdf:reifies</code> property. A triple where the predicate is
    <code>rdf:reifies</code> and the object is therefore a proposition, is
    a <dfn>reifying triple</dfn>, and its subject is called a <dfn>reifier</dfn>.
```

spec/index.html
```diff
@@ -1243,6 +1250,8 @@ <h2>RDFS Interpretations</h2>
     <em>Ext</em>ension in I) from IC to the set of subsets of IR.</p><p> A class may have an
     empty class extension. Two different classes can have the same class extension.
     The class extension of <code>rdfs:Class</code> contains the class <code>rdfs:Class</code>.</p>
+    
+    <p>RDFS also introduces the class <code>rdfs:Proposition</code>, which denotes a resource whose extension is exactly the <a>set of propositions</a> as defined in [[[#simple_entailment_properties]]]. The denotation of <code>rdfs:Proposition</code> also includes the range of the <code>rdf:reifies</code> property. A triple where the predicate is <code>rdf:reifies</code> and the object is therefore a proposition, is a <dfn>reifying triple</dfn>, and its subject is called a <dfn>reifier</dfn>. 
 
   <p> An <dfn>RDFS interpretation</dfn> (<strong>recognizing D</strong>) is an <a>RDF interpretation</a> (recognizing D) I
     which <a>satisfies</a> the semantic conditions in the following table, and all the triples in the subsequent table of RDFS axiomatic triples.</p>
```
TallTed, line 1257:
```suggestion
  <p> An <dfn>RDFS interpretation</dfn> (<strong>recognizing D</strong>) is an
  <a>RDF interpretation</a> (recognizing D)&nbsp;I which <a>satisfies</a> the
  semantic conditions in the following table, and all the triples in the
  subsequent table of RDFS axiomatic triples.</p>
```

TallTed (changes requested):
Adding a lot of line breaks and `&nbsp;`. Small tweaks to some text.

spec/index.html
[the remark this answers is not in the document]
niklasl: @TallTed I think your improvement was on the "before" (i.e. doesn't contain my rewording and Concepts-links), so here is my suggestion with shorter lines (and fixed indent) per your advice:
```suggestion
    <p>RDFS also introduces the class <code>rdfs:Proposition</code>, which denotes
    a resource whose extension is exactly the <a>set of propositions</a> as defined
    in [[[#simple_entailment_properties]]]. This class is also the range of the
    <code>rdf:reifies</code> property. A triple where the predicate is
    <code>rdf:reifies</code> is called a
    <a data-cite="RDF12-CONCEPTS#dfn-reifying-triple">reifying triple</a>,
    and its subject is called a
    <a data-cite="RDF12-CONCEPTS#dfn-reifier">reifier</a>. Since its predicate is
    <code>rdf:reifies</code>, the object therefore denotes a
    <a data-cite="RDF12-CONCEPTS#dfn-proposition">proposition</a>.
 ```
 (Of course the detailed content changes are harder to read here; see my oneline-change above for clarity.)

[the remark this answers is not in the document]
doerthe: Do we need the last sentence if we already say that the range of reifies is proposition?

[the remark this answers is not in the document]
doerthe: I agree. We can still state that the range of `rdf:reifies` is `rdfs:Proposition` as that IS part of the semantics.

doerthe:
I am not @franconi but I dare to answer :)

> **In section 5.3** the added text has helped me understand, but there is still room for improvement I guess. In particular I find the following confusing:
> 
> > We define the set of facts in an interpretation as follows:
> > [...]
> > [...] a fact in an interpretation is the proposition corresponding to the denotation of an asserted triple in the interpretation.
> > Given a blank node mapping, we define the set of facts asserted by a graph in an interpretation as follows:
> > [...]
> > an asserted fact in a graph is the proposition corresponding to the denotation of a triple in the graph.
> 
> I find it quite difficult to spot the difference between these two sentences. In what respect is this difference important?
> 
I agree with you that this is quite difficult to grasp (I already suggested to add at least the letters used in the definitions to help the reader digest). So the difference is:

F(I):  facts in an interpretation = triples (interpreted as resources) which are true according to an interpretation. So, someone gives me I and I now simple have a set of triples which is true there

FEXT(G,I,A): facts asserted by a graph under I and some blank node mapping A = the denotations (= interpretations of triples as resources) of the triples listed in the graph.

So the first part refers to the denotations of all triples which are true according to the interpretation (no graph needed) and the second part refers to the denotation of all triples occurring in the graph.

If I is actually a model of G, then we have for some A that FEXT(G,I,A) is a subset of F(I).

> **In section 9** the [added paragraph](https://github.com/w3c/rdf-semantics/pull/156#issuecomment-3254599687) IMO does not help comprehension.
> 
> > RDFS also introduces the class rdfs:Proposition , which denotes a resource whose extension is exactly the[ set of propositions ](https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/156/6639a2a...franconi:39cde88.html#dfn-set-of-propositions)as defined in 5.3 Properties of simple entailment and satisfiability . The denotation of rdfs:Proposition also includes the range of the rdf:reifies property. A triple where the predicate is rdf:reifies and the object is therefore a proposition, is a reifying triple , and its subject is called a reifier .
> 
> This re-states stuff stated in other specs, but doesn't add anything new.

You are right for the last part (and you agree there with @pchampin), I think it is still interesting to see that the set of propositions from simple entailment  becomes `rdfs:Proposition`.  Do you consider that part helpful or still not so much?

> 
> **In section F** there seems to be a typo: the surname of Jean-Yves Rossi is written in caps as `Jean-Yves ROSSI`.
> 
> **In summary** I'm increasingly leaning towards the opinion expressed repeatedly by you and @pfps that the Semantics document provides all it should provide (modulo some clarifications, naturally). A tighter definition of reification should be pursued in RDF Concepts and RDF Schema.

+1

spec/index.html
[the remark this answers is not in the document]
niklasl: Good points @pchampin. I understand, and agree (also with @doerthe's comment).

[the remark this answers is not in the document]
rat10: Can you add another sentence to explain how these propositions, which are expressed as triple terms and therefore very much resemble facts, correspond to facts, asserted or not? Or is this request confused, and confusing?

doerthe:
> Okay, that is the difference between facts and asserted facts that has puzzled me before already. I still wonder what the importance of F(I) is, as in practice it seems that all we are concerned about is the set of asserted facts. I always find it much easier to understand a formal text like this one if I understand for each part the motivation, the purpose, the way in which it interacts with the other parts and why/how it is needed.

You always need an interpretation to judge whether or not asserted triples are true. The idea behind our definitions is now  that that someone gives us the set of all triples (it is more complicated, the resources our triples denote, but let's not go there) which are true according to the interpretation (this is most likely an infinite set), this is the set F(I). 

If we now want to know, whether our asserted triples are true, we do not need to apply the interpretation function IEXT on the triples but we can instead simply check whether the triple (more precisely its denotation) is in the set of facts (the statements of which we know that they are true).  If all the triples in the graph (the asserted triple) are also in the set of facts (the true triples), then the graph is true according to the interpretation.

I am not sure that helps, but at least I tried and I will answer to questions :)

spec/index.html
pchampin: Plus, as you pointed out elsewhere, the definition of "set of fact" does not even refer to any graph, so the notion of asserted triple does not even apply. :)

rat10:
> You always need an interpretation to judge whether or not asserted triples are true. The idea behind our definitions is now that that someone gives us the set of all triples (it is more complicated, the resources our triples denote, but let's not go there) which are true according to the interpretation (this is most likely an infinite set), this is the set F(I).

So far I seem to follow.

> If we now want to know, whether our asserted triples are true, we do not need to apply the interpretation function IEXT on the triples but we can instead simply check whether the triple (more precisely its denotation) is in the set of facts (the statements of which we know that they are true). If all the triples in the graph (the asserted triples) are also in the set of facts (the true triples), then the graph is true according to the interpretation.
> 
> I am not sure that helps, but at least I tried and I will answer to questions :)

So this is a kind of optimization device, a definition that helps checking if a graph is true? Was that asked for? Is it helpful in practice? (I imagine that one could implement this without needing the RDF Semantics to specify it)

PR #157, opened 2025-09-16
Fixed skolem text with simple entailment

franconi:
Whenever the words "entails/entailment" appear in the Skolemization section then they should always be "simply entails/simple entailment". You can not even imagine the fight I had to do in the RDF 1.1 WG when the SPARQL OWL entailment regime was introduced to convince Birte and Bijan that this was the case. I lost the fight, and the SPARQL OWL entailment regime is simply semantically wrong and it has never been applied in real tools.
Connected with issue https://github.com/w3c/rdf-concepts/issues/239

***
<a href="https://pr-preview.s3.amazonaws.com/franconi/rdf-semantics-fork-enrico/pull/157.html" title="Last updated on Sep 16, 2025, 2:33 PM UTC (eec41d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/157/d2292b4...franconi:eec41d1.html" title="Last updated on Sep 16, 2025, 2:33 PM UTC (eec41d1)">Diff</a>


pchampin:
Out of curiosity, why do these property only hold for simple entailment?? Are there counter examples where they do not hold in RDF entailmenet or RDFS entailment??

spec/index.html
```diff
@@ -780,7 +780,7 @@ <h2>Skolemization</h2>
...
+    Nevertheless, they are in a strong sense almost interchangeable <a data-cite="RDF12-SEMANTICS#simple">RDF simple interpretations</a>, as shown the next two properties. The third property means that even when conclusions are drawn with <a data-cite="RDF12-SEMANTICS#simple">simple entailment</a> from the skolemized graph which do contain the new vocabulary, these will exactly mirror what could have been derived with <a data-cite="RDF12-SEMANTICS#simple">simple entailment</a> from the original graph with the original blank nodes in place. The replacement of blank nodes by IRIs does not effectively alter what can be validly derived with <a data-cite="RDF12-SEMANTICS#simple">simple entailment</a> from the graph, other than by giving new <a>name</a>s to what were formerly anonymous entities. The fourth property, which is a consequence of the third, clearly shows that in some sense  a skolemization of G can "stand in for" G as far as <a data-cite="RDF12-SEMANTICS#simple">simple entailments</a> are concerned. Using sk(G) instead of G will not affect any <a data-cite="RDF12-SEMANTICS#simple">simple entailments</a> which do not involve the new skolem vocabulary.  </p>
```
TallTed, line 783: There are several changes below, moving words around and putting at least one plural inside an `<a>` tag. I don't believe I've changed any intended meaning with any of these changes. GitHub won't highlight these changes, because I've also broken the EXTREMELY long original line into a bunch of shorter lines. Please review with your diff tool of choice.
```suggestion
    Nevertheless, they have in a strong sense almost interchangeable
    <a data-cite="RDF12-SEMANTICS#simple">RDF simple interpretations</a>,
    as shown by the next two properties. The third property means that even
    when conclusions which do contain the new vocabulary are drawn with
    <a data-cite="RDF12-SEMANTICS#simple">simple entailment</a> from the
    skolemized graph, these will exactly mirror what could have been derived with
    <a data-cite="RDF12-SEMANTICS#simple">simple entailment</a> from the original
    graph with the original blank nodes in place. The replacement of blank nodes
    with IRIs does not effectively alter what can be validly derived from the
    graph with <a data-cite="RDF12-SEMANTICS#simple">simple entailment</a>,
    other than by giving new <a>names</a> to what were formerly anonymous
    entities. The fourth property, which is a consequence of the third, clearly
    shows that in some sense, a skolemization of G can "stand in for" G as far
    as <a data-cite="RDF12-SEMANTICS#simple">simple entailments</a> are
    concerned. Using sk(G) instead of G will not affect any
    <a data-cite="RDF12-SEMANTICS#simple">simple entailments</a> which do not
    involve the new skolem vocabulary.</p>
```

pfps:
> > Out of curiosity, why do these property only hold for simple entailment?? Are there counter examples where they do not hold in RDF entailmenet or RDFS entailment??
> 
> It holds also for RDF and RDFS entailments - it basically holds for any logic extension to simple RDF enjoying the minimal model property. It does not hold for OWL. But I think that it is fine to mention only simple entailment, since RDF and RDFS entailments can be reduced to simple entailment with model preservation (namely, the completion rules).

Doesn't this also "hold" for OWL, in the sense that skolemization in OWL will either produce the entailment or create something that is not OWL?

franconi:
> Doesn't this also "hold" for OWL, in the sense that skolemization in OWL will either produce the entailment or create something that is not OWL?

First, I am not sure that at least a case producing the entailment exists (do you have an example?).
Second, skolemising OWL generates I'll-formed OWL most/all of the times, so the theorems would be void.
Third, your belief is exactly what lead Bijan and Birte to define SPARQL under OWL entailment a monster, since bnodes in the query are skolemised - an horror for the description logics community...

pchampin:
thanks @pfps @franconi for the clarification :)

spec/index.html
TallTed: submitted as PR #159

PR #168, opened 2025-11-11
added non-well foundedness note

franconi:
Added a technical note discussing why non-well founded interpretations are harmless in RDF

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/pull/168.html" title="Last updated on Dec 18, 2025, 9:15 AM UTC (1f4793e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-semantics/168/294ddeb...1f4793e.html" title="Last updated on Dec 18, 2025, 9:15 AM UTC (1f4793e)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1513,6 +1513,40 @@
 
     <p> aaa <code>rdf:type rdfs:Resource .</code></p>
 
+	<p class="technote"> 
+		The RDFS semantic condition on the triple term mapping IT induces 
+		unwanted <i>non-well founded models</i> for propositions.
+		
+		For example, the asserted triple 
+		
+		<p></p>&lt;&lt;( :a :b :c )&gt;&gt; :p :o.</p>
+	  
+		is satisfied in the interpretation where IT maps &lt;x,y,z&gt; to x, 
+		and x is an instance of the denotation of rdfs:Proposition. 
+	  
+		In this interpretation, there is a proposition about a relation involving the proposition itself 
+		-- for example it could be intended as the proposition "the barber denies this very proposition",
+		leading to a self referential paradoxe. 
+	  
+	  However, it should be noted that: 
+	  
+		<ol>
+			<li> interpretations like the above do not induce any formal paradox, 
+				since in RDF the notion of proposition is very weak, and most importantly, </li>
+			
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1515,6 +1515,31 @@ <h3>RDFS entailment</h3>
 
     <p>where aaa is an IRI, are true in all RDFS interpretations.</p>
 
+	<p class="technote"> 
+		The RDFS semantic condition on the triple term mapping IT induces 
+		unwanted <i>non-well founded models</i> for propositions.
```
pfps:
```suggestion
		potentially unwanted <i>non-well founded models</i> for propositions.		
```

pfps (approved):

TallTed (dismissed):
Here, again, we have formatting (most line breaks) to make the *source* clearer to reviewers. Similar formatting should be applied to the content being rendered to readers of the ReSpec *output*, to similarly increase comprehension. I have included such in my change requests on this PR. Use your favorite DIFF tool to reveal what GitHub does not.

pchampin (changes requested):
I have some issues with the very beginning of this note (and with its location), but the rest of the note looks good.

franconi:
Ok ok, I agree.
But still I would keep the notion of proposition in this text.

doerthe:
I missed the  whole discussion here. I am not sure that we need  the remark at all (but I acn be convinced). Yes we have some possible self-reference in what you call "intended meaning", but I somehow do not see why we want to mention it.

franconi:
> I missed the whole discussion here. I am not sure that we need the remark at all (but I can be convinced). Yes we have some possible self-reference in what you call "intended meaning", but I somehow do not see why we want to mention it.

We have to mention this fact in RDF-Semantics since the current _normative_ semantics does not fully capture the intended meaning of a triple term as a _proposition_. We also observe that the unintended models induced by the normative semantics have no impact on the behaviour of simple, RDF, and RDFS entailments.
It is still debatable where in the document this would better fit.

franconi:
@pchampin, is it better now?

pchampin (approved):
Looks good now, thanks.

